PR #58, opened 2021-12-01
try-with-resources for encoder/decoder

graemerocher:
Improvements encoder/decoder APIs to use try with resources

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/Decoder.java
+++ b/serde-api/src/main/java/io/micronaut/serde/Decoder.java
@@ -212,6 +212,10 @@ public interface Decoder extends AutoCloseable {
      */
     void finishStructure() throws IOException;
 
+    /**
+     * Finalize the current structure. Equivalent to calling {@link #finishStructure()}.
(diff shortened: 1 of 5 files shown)


serde-api/src/main/java/io/micronaut/serde/Decoder.java
```diff
@@ -212,6 +212,11 @@ default Decoder decodeObject() throws IOException {
      */
     void finishStructure() throws IOException;
 
+    @Override
+    default void close() throws IOException {
```
yawkat, line 216: should document that this is equivalent to finishStructure

serde-api/src/main/java/io/micronaut/serde/Encoder.java
```diff
@@ -50,6 +50,11 @@
      */
     void finishStructure() throws IOException;
 
+    @Override
+    default void close() throws IOException {
```
yawkat, line 54: should document that this is equivalent to finishStructure

yawkat (approved):

PR #59, opened 2021-12-02
Support JsonBackReference

graemerocher:


serde-api/src/main/java/io/micronaut/serde/serializers/ObjectSerializer.java
```diff
@@ -274,8 +309,13 @@ public final void serialize(
                 }
             }
             childEncoder.finishStructure();
+        } catch (StackOverflowError e) {
+            throw new SerdeException("Infinite recursion serializing type: " + type.getType()
```
yawkat, line 313: is it really safe to just catch SOE? could also just have a depth counter, or ideally catch this at compile time

yawkat (approved):

serde-api/src/main/java/io/micronaut/serde/serializers/ObjectSerializer.java
graemerocher: This is what Jackson appears to do

yawkat: i guess it's not so bad since it cant be caused by a user, only by programmer error

dstepanov (approved):

PR #65, opened 2021-12-03
Implement currentPath for jsonp and bson encoders

### LINKED ISSUE #60: Implement `Encoder.currentPath` for BSON
### Feature description

Currently returns a blank path, should return the path for debugging purposes

### LINKED ISSUE #61: Implement `currentPath` for JSON-P
### Feature description

Currently returns a blank path

yawkat:
Path is calculated lazily because it's only needed for error messages. Using a reference to the parent encoder keeps memory use relatively low.

Resolves #60
Resolves #61

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/Encoder.java
+++ b/serde-api/src/main/java/io/micronaut/serde/Encoder.java
@@ -150,7 +150,7 @@ public interface Encoder extends AutoCloseable {
     void encodeNull() throws IOException;
 
     /**
-     * Return an anaysis of the current path.
+     * Return an analysis of the current path.
      *
      * @return The current path if known
      */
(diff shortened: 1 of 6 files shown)


graemerocher:
Doesn't seem to handle current array index?

yawkat:
index is annoying to do reliably, because there's many encode methods, and bson has additional external writes through `getBsonWriter` that we can't see. But I will implement it best-effort

yawkat:
hm, now failing with `Cannot auto-detect encoding, not enough chars` in the jsonp reactive parser. Not sure what change triggered it, or if it's spurious. It's weird that itd try detecting the charset anyway, since json is specified to always be utf-8...

yawkat:
fixed by rebase, ready for review now @graemerocher

graemerocher (approved):

PR #76, opened 2021-12-10
Serialization of AWS events

sdelamo:
Includes a module to provide Serde configuration for [AWS Lambda Events](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events) POJOs.


graemerocher:
@sdelamo the latest PR from @dstepanov might help with the AWS case https://github.com/micronaut-projects/micronaut-serialization/pull/78

dstepanov:
@sdelamo What problem do you have?

sdelamo:
@dstepanov This PR contains a test annotated with `PendingFeature`  to test serialization into a `SQSEvent` a POJO distributed by AWS Lambda events library. They ship POJOs and Mixins. 

I would like to be able to apply serialize to [SQSEvent](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SQSEvent.java) leveraging [SQSEventMixin](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/events/mixins/SQSEventMixin.java)

Do you know how could I achieve that?

sdelamo:
@dstepanov do you have any feedback about this?

dstepanov:
It might be possible to use `@SerdeImport` with a custom naming strategy that converts `records` to `Records` etc. The best way would be to support something like mixins with: `@SerdeImport(value=Xyz.class, mixin=XyzMixin.class)` WDYT @graemerocher? I think we already have some kind of annotation hierarchy, mixin would contribute the first place matched by the name of field/property/constructor parameter.

graemerocher:
Sounds good

graemerocher:
I will take a look at this

graemerocher:
PR submitted with a solution for this https://github.com/micronaut-projects/micronaut-serialization/pull/102

graemerocher:
FWIW I think this code belongs in the AWS module not here

sdelamo:
I added the mixin as in @yawkat PR but the test is still failing. 

@yawkat any idea?

yawkat:
im not sure, but the mixin being in a different jar that isnt covered by the annotation processor could be the problem

graemerocher:
The reason this is failing is because AWS re-packages Jackson and the byte code the annotations are `com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.annotation.JsonProperty` 😱

yawkat:
🤦

sdelamo:
> The reason this is failing is because AWS re-packages Jackson and the byte code the annotations are `com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.annotation.JsonProperty` 😱

Should an annotation mapper which maps `@JsonProperty` to `com.fasterxml.jackson.annotation.JsonProperty` help?

graemerocher:
See https://github.com/micronaut-projects/micronaut-serialization/pull/108

sdelamo:
@yawkat and @graemerocher The `mixin` works great. 

`aws-events` library uses joda and it has fields of type `org.joda.time.DateTime`. 

They ship with two `com.fasterxml.jackson.databind.module.SimpleModule`

https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/events/modules

How could we handle this?

graemerocher:
Add `Serializer` and `Deserializer`  beans (or a single bean of type `Serde`) for `DateTime`

graemerocher:
I committed what you need to add

sdelamo:
@graemerocher 

> FWIW I think this code belongs in the AWS module not here

I have moved it to aws https://github.com/micronaut-projects/micronaut-aws/pull/1264

PR #94, opened 2022-01-04
Move bean classes to support module

graemerocher:
This minimizes the `serde-api` module to include only annotations and interfaces


serde-api/build.gradle.kts
```diff
@@ -4,22 +4,9 @@ plugins {
 
 dependencies {
     annotationProcessor(mn.micronaut.inject.java)
-    annotationProcessor(mn.micronaut.inject.java)
-    annotationProcessor("io.micronaut.docs:micronaut-docs-asciidoc-config-props:2.0.0.RC1")
+
     api(mn.micronaut.context)
```
yawkat, line 8: is context needed? I see `io.micronaut.core.annotation.Indexed`, for that I guess?

yawkat (approved):

serde-api/build.gradle.kts
graemerocher: it is needed for https://github.com/micronaut-projects/micronaut-serialization/blob/master/serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java

sdelamo (approved):

PR #99, opened 2022-01-05
build: align build with other modules

sdelamo:
I've changed the build to be closer to what we do in other modules:

- bump up micronaut to 3.2.4
- bump up micronautDocsVersion to 2.0.0
- Set projectGroup in gradle.properties
- replace docs-examples by convention test-suite modules
- remove serde-build-logic.
- remove serde-module apply io.micronaut.build.internal.module instead. serde-build-logic was using io.micronaut.build.internal.base-module not sure why. We use module not base-module in the rest of the modules.


graemerocher:
Please revert the change that moves the doc examples to a `test-suite` directory. These are not test suites but documentation examples and many other repos use this pattern https://github.com/micronaut-projects/micronaut-data/tree/master/doc-examples

sdelamo:
We use heavily `test-suite` to provide documentation snippets for other modules. Maybe `data` should be migrated to test-suite to follow the convention of the rest of the modules.

graemerocher:
grpc, gcp, pulsar, data, azure, r2dbc, serialization all use the doc-examples convention https://github.com/search?q=org%3Amicronaut-projects+doc-examples&type=code

only core, aws and security use https://github.com/search?q=org%3Amicronaut-projects+test-suite&type=code as far as I can see

In addition these are not test suites, but projects specifically for documenting the project, so I see no issue or inconsistency here.

sdelamo:
I've reverted the changed and left `docs-examples`. 

`email`, `aws`, `views`, `core`, `elasticsearch`, `problem-json`, `graphql`, `spring` use `test-suite`.

Please, note that the [`bom` plugin excludes projects with `test-suite`](https://github.com/micronaut-projects/micronaut-build/blob/master/src/main/groovy/io/micronaut/build/MicronautBomPlugin.java#L82). We don't have any special treatment for `docs-examples` as far as I know. . 

I think it would be simpler just to stick with one convention and use `test-suite`.

graemerocher:
* `graphql` uses doc-examples https://github.com/micronaut-projects/micronaut-graphql/blob/master/settings.gradle
* `spring` and `elasticsearch` don't t use it for anything else besides tests
* the remaining modules you mention seem to be ones you created or maintain :)

I do agree that the bom plugin needs to include it although I think the BOM plugin should only include a project if it has the module plugin applied then the hard coded exclusions wouldn't be needed. Thoughts @melix ?

melix:
> the BOM plugin should only include a project if it has the module plugin applied

Eventually it should be something like that. It would work however today even if you don't. There are also some ordering issues which make this a bit unreliable atm.

PR #102, opened 2022-01-05
support for mixins when importing external types

graemerocher:
Adds the ability to apply mixins to customise mapping when importing external types


doc-examples/example-java/build.gradle.kts
```diff
@@ -22,3 +22,10 @@ dependencies {
 application {
     mainClass.set("example.Application")
 }
+
+//tasks {
```
yawkat, line 26: did you leave this intentionally for docs?

yawkat (approved):

doc-examples/example-java/build.gradle.kts
graemerocher: No, just in case I needed it again I left it there :)

PR #107, opened 2022-01-07
Fix issues identified supporting problem/json

graemerocher:
* If no introspection is present fallback to trying to resolve based on the runtime type
* `@Deserializable` should be annotated with `@Introspected`
* Allow deserializing raw maps

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/deserializers/CoreDeserializers.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/deserializers/CoreDeserializers.java
@@ -210,39 +210,54 @@ public class CoreDeserializers {
         @Override
         default M deserializeNonNull(Decoder decoder, Deserializer.DecoderContext decoderContext, Argument<? super M> type) throws IOException {
             final Argument<?>[] generics = type.getTypeParameters();
+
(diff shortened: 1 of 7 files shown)


serde-support/src/main/java/io/micronaut/serde/support/deserializers/CoreDeserializers.java
```diff
@@ -210,39 +210,54 @@
...
+                    return map;
                 } else {
-                    try {
-                        k = decoderContext.getConversionService().convertRequired(
-                                key,
-                                keyType
-                        );
-                    } catch (ConversionErrorException e) {
-                        throw new SerdeException("Error converting Map key [" + key + "] to target type [" + keyType + "]: " + e.getMessage(), e);
+                    return getDefaultValue();
```
yawkat, line 224: should be an error imo

yawkat (approved):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/CoreDeserializers.java
graemerocher: Agreed, make it an error

PR #108, opened 2022-01-07
Support for AWS mixins

graemerocher:
This is a draft PR, it is broken right now because of a bug in dynamic annotation of inner classes


graemerocher:
PR to core to fix the issue with SQSMessageMixin https://github.com/micronaut-projects/micronaut-core/pull/6721

PR #116, opened 2022-01-17
Add decoder context and argument type in `getDefault` method

dstepanov:
BTW it might be a good idea to change the order of the parameters in `createSpecific` to have the same order as in `deserialize`: context and then the type, in the same way, `serialize` can have context, type, and then value.


graemerocher:
makes sense, do you want to make that change?

dstepanov:
I will

PR #122, opened 2022-01-26
fix java.lang.IllegalStateException: Read only 0 bytes!

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-data/issues/1295

### DIFF
--- a/serde-bson/src/main/java/io/micronaut/serde/bson/AbstractBsonMapper.java
+++ b/serde-bson/src/main/java/io/micronaut/serde/bson/AbstractBsonMapper.java
@@ -127,6 +127,15 @@ public abstract class AbstractBsonMapper implements ObjectMapper {
                     return writeValueToTree(o);
                 }
             }
+
+            @Override
+            protected JsonNode parseOne(byte[] remaining) throws IOException {
+                try (BsonReader bsonReader = createBsonReader(ByteBuffer.wrap(remaining))) {
+                    final BsonReaderDecoder decoder = new BsonReaderDecoder(bsonReader);
+                    final Object o = decoder.decodeArbitrary();
+                    return writeValueToTree(o);
+                }
+            }
         };
     }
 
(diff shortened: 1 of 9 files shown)


dstepanov (approved):

serde-bson/src/main/java/io/micronaut/serde/bson/AbstractBsonMapper.java
```diff
@@ -176,10 +176,7 @@ private ByteBuffer toByteBuffer(InputStream inputStream) throws IOException {
         ReadableByteChannel channel = Channels.newChannel(inputStream);
         while (true) {
             int read = channel.read(byteBuffer);
-            if (read == 0) {
-                throw new IllegalStateException("Read only 0 bytes!");
-            }
-            if (read == -1) {
+            if (read <= 0) {
```
yawkat, line 179: is this the right thing to do? returning 0 from read is technically invalid, but it doesnt have to mean EOF necessarily

graemerocher: I'm not sure, but without this check it loops forever since calling read again and again just always returns 0 if there is no more data

yawkat: ok, maybe it's a quirk of the bson streams then

yawkat (approved):

graemerocher:
I don't think so, the type that is returning 0 is `ByteArrayInputStream` which is a JDK class

yawkat:
oh i see. because the buffer is exactly the size of the input stream, the buffer has no more space in the final `read` call, in which case it is valid to return 0.

But that means that this code doesn't work (the previous one didn't either) if `InputStream.available` is wrong. `InputStream.available` is an optional method, and it looks like users can supply their own streams, so we should fix this

graemerocher:
yeah I am not sure relying on `available()` is the correct thing to do here. Seems we should create a reasonable size buffer and then resize if necessary 🤔

yawkat:
just copy to a ByteArrayOutputStream and then ByteBuffer.wrap the returned byte array, i guess

graemerocher:
yeah probably for the best

yawkat:
lgtm

PR #123, opened 2022-01-28
inclusion of io.micronaut introspections package configurable

sdelamo:


graemerocher:
@sdelamo I simplified the change a bit since I don't think it is necessary to add another config option you can just set the existing one to the list of packages you want to include or an empty value if you don't want to include any

sdelamo:
I intentionally added another option because I think it would be easy for users to remove `io.micronaut` as a serdeable package by mistake. 

If a user sets : 

```yaml
micronaut:
    serde:
        includedIntrospection-packages: 
          - 'com.myacompany' 

```

then `io.micronaut` will not longer be used. 

They will need to set: 

```yaml
micronaut:
    serde:
        includedIntrospection-packages: 
          - 'io.micronaut'       
          - 'com.myacompany' 

```

sdelamo:
Moreover, the current Configuration Reference documentation does not provide the uesr information that `io.micronaut` is used as default value.

See: https://github.com/micronaut-projects/micronaut-serialization/issues/126

PR #130, opened 2022-02-02
add DeserializationConfiguration default value

sdelamo:
Javadoc says it defaults to 100 but the bean of type `DeserializationConfiguration` does not have any default value.


graemerocher (approved):

sdelamo:
once this get merged I think it would be good to release a new Milestone. Without this fix users, will need to define `micronaut.serde.deserialization.array-size-threshold` in certain scenarios https://github.com/micronaut-advocacy/micronaut-live-newsletter/pull/8/files#diff-cfaac170cb673164b4d30a6ae0f20631762be4126821b99e86a4421f09bf4fda

graemerocher:
can you fix the checkstyle

PR #211, opened 2022-05-18
Fix subtyping serialization

### LINKED ISSUE #176: JsonError._embedded serialization broken with micronaut 3.4.0
### Issue description

In 3.4.0, the `Resource` class became `@Introspected`. Before this change, the `OptionalMultiValuesSerializer` would fall back on the `ObjectSerializer`, which in turn would use a `RuntimeTypeSerializer` which would delegate to the right subtype for serialization.

However after the change, `ObjectSerializer` will simply specialize to the serializer for `Resource`, which will ignore properties of subtypes of `Resource` (e.g. the `message` of a `JsonError`). This manifests in the `test read / write JsonError` test case failing.

Not sure how to fix this. It seems to me like the addition of `@Introspected` should not change serialization behavior this way.

### LINKED ISSUE #204: Fails to serialise Parent->Child objects with JsonTypeInfo, JsonSubTypes
### Expected Behavior

```
{"type":"SecondChild","fieldOfSecond":1,"sibling":{"name":"first", "type": "FirstChild", "fieldOfFirst": true},"name":"second"}
```

### Actual Behaviour

```
{"type":"SecondChild","fieldOfSecond":1,"sibling":{"name":"first"},"name":"second"}
```

### Steps To Reproduce

Given this Parent
```
@JsonTypeInfo(
    use = JsonTypeInfo.Id.NAME,
    property = "type")
@JsonSubTypes({
    @JsonSubTypes.Type(value = FirstChild.class, name = "FirstChild"),
    @JsonSubTypes.Type(value = SecondChild.class, name = "SecondChild")
})
@Serdeable
public class Parent {
```
Having 2 child classes
```
@Serdeable
public class FirstChild extends Parent {
    private boolean fieldOfFirst;
```
and
```
@Serdeable
public class SecondChild extends Parent {
    private int fieldOfSecond;
    private Parent sibling;
```
Then on serialising:
```
FirstChild first = (FirstChild) new FirstChild().setFieldOfFirst(true).setName("first");
SecondChild secondChild = (SecondChild) new SecondChild().setFieldOfSecond(1).setSibling(first).setName("second");
String json = mapper.writeValueAsString(secondChild);
```

Note that marking Parent as abstract makes it even worse since it tries to instantiate it.

### Environment Information

- MacOs
- Java 11.0.14-tem

### Example Application

https://github.com/bowika/micronaut-serialization-bugreport

### Version

3.4.3

graemerocher:
Use the runtime type if subtyping is involved or the type is abstract. Fixes #204

Also fixes #176

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
-projectVersion=1.0.2-SNAPSHOT
+projectVersion=1.1.0-SNAPSHOT
 projectGroup=io.micronaut.serde
 micronautDocsVersion=2.0.0
-micronautVersion=3.3.3
+micronautVersion=3.4.4
 micronautTestVersion=3.0.5
 groovyVersion=3.0.9
 spockVersion=2.0-groovy-3.0
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonTypeInfoSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonTypeInfoSpec.groovy
@@ -461,4 +461,58 @@ class Cat extends Animal {
         cleanup:
         context.close()
     }
+
+    void "test nested subtypes"() {
+        given:
+        def context = buildContext("""
+package subtypes;
+
+import com.fasterxml.jackson.annotation.*;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.serde.annotation.Serdeable;
+
+@JsonTypeInfo(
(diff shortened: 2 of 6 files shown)


yawkat:
What about #176

graemerocher:
will add a test

yawkat:
ah i guess resource is abstract

graemerocher:
I upgraded the project to 3.4.4 and the test in `CoreTypeSerdeSpec` passes

yawkat (approved):

dstepanov (approved):

PR #247, opened 2022-07-20
Fix null handling in serde-imported classes with object arrays

### LINKED ISSUE #245: Serialization of an array of external classes containing a null fails
### Expected Behavior

I'm calling a 3rd party library and serializing the results.

I use `@SerdeImport` to enable serialization on the class (and a mixin to reveal null values).

However, the 3rd party library can have nulls in an array field in the class, and when this happens, serialization fails.

It looks like it tries to call methods on the null object during serialization of the array.  I have tried both 1.1.0 and 1.1.1-SNAPSHOT of the serialization array

### Actual Behaviour

_No response_

### Steps To Reproduce

- Clone the example repo
- Run it
- Hit http://localhost:8080/good and see json
- Hit http://localhost:8080/bad and see an error in the logs

### Environment Information

OSX
Java 11

### Example Application

https://github.com/bloidonia/nested_serde

### Version

3.5.3 (fails on 3.6.0-SNAPSHOT too)

timyates:
Fixes #245

### DIFF
--- a/serde-jackson/build.gradle.kts
+++ b/serde-jackson/build.gradle.kts
@@ -23,6 +23,7 @@ dependencies {
     if (!JavaVersion.current().isJava9Compatible()) {
         testImplementation(files(org.gradle.internal.jvm.Jvm.current().toolsJar))
     }
+    testImplementation("one.microstream:microstream-storage-restclient:07.00.00-MS-GA")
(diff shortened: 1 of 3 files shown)


dstepanov (approved):

timyates:
Hold off merging this, I want to confirm it fixes the error in microstream

PR #251, opened 2022-07-26
Allow access to a default object mapper 

graemerocher:
In some cases static access is still necessary, and we don't really have a solution for that. This is a proposal to add a static `ObjectMapper` implementation to replace cases where people are manually instantiating Jackson's `ObjectMapper` because access to the context is not viable


serde-api/src/main/java/io/micronaut/serde/ObjectMappers.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ *
+ * @author graemerocher
+ */
+final class ObjectMappers {
+    private static ObjectMapper defaultObjectMapper;
+    private static BeanContext beanContext;
+
+    static ObjectMapper resolveDefault() {
```
yawkat, line 38: not thread safe

yawkat (changes requested):
we could also put something like this on JsonMapper

serde-api/src/main/java/io/micronaut/serde/ObjectMappers.java
```diff
@@ -0,0 +1,108 @@
...
+import io.micronaut.context.env.Environment;
+import io.micronaut.inject.BeanDefinitionReference;
+
+/**
+ *
+ * @author graemerocher
+ * @since 1.2.2
+ */
+final class ObjectMappers {
+    private static ObjectMapper defaultObjectMapper;
```
yawkat, line 36: needs to be volatile

serde-api/src/main/java/io/micronaut/serde/ObjectMappers.java
```diff
@@ -0,0 +1,108 @@
...
+    private static BeanContext beanContext;
```
yawkat, line 37: this too

yawkat (approved):

PR #290, opened 2022-10-23
test: annotate failing test with @PendingFeature

sdelamo:
Test fails with: 

```
  Condition failed with Exception:test io.micronaut.serde.jackson.object.ObjectSerdeSpec
  
  writeJson(jsonMapper, beanUnderTest) == '{"foo":"bar"}'
  |         |           |
  |         |           <example.Test@190728b4 foo=bar>
  |         <io.micronaut.serde.jackson.JacksonJsonMapper@4e6f9498 registry=io.micronaut.serde.support.DefaultSerdeRegistry@4a8375f9 deserializationConfig=io.micronaut.json.JsonStreamConfig@275581aa treeCodec=io.micronaut.jackson.core.tree.JsonNodeTreeCodec@da1056e objectCodecImpl=io.micronaut.serde.jackson.JacksonJsonMapper$ObjectCodecImpl@2c634771 view=class java.lang.Object encoderContext=io.micronaut.serde.support.DefaultSerdeRegistry$1@175b4fae decoderContext=io.micronaut.serde.support.DefaultSerdeRegistry$2@1c88d43e>
  java.lang.ClassCastException: class java.util.Optional cannot be cast to class java.lang.String (java.util.Optional and java.lang.String are in module java.base of loader 'bootstrap')
        at io.micronaut.serde.support.serializers.CoreSerializers$2.isEmpty(CoreSerializers.java:63)
        at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serialize(CustomizedObjectSerializer.java:94)
        at io.micronaut.serde.jackson.JacksonJsonMapper.writeValue(JacksonJsonMapper.java:108)
        at io.micronaut.serde.jackson.JacksonJsonMapper.writeValue0(JacksonJsonMapper.java:100)
        at io.micronaut.serde.jackson.JacksonJsonMapper.writeValue0(JacksonJsonMapper.java:94)
        at io.micronaut.serde.jackson.JacksonJsonMapper.writeValueAsBytes(JacksonJsonMapper.java:189)
        at io.micronaut.serde.jackson.JsonSpec$Trait$Helper.writeJson(JsonSpec.groovy:7)
        at io.micronaut.serde.jackson.object.ObjectSerdeSpec.optional nullable mix(ObjectSerdeSpec.groovy:1927)
      at io.micronaut.serde.jackson.object.ObjectSerdeSpec.optional nullable mix(ObjectSerdeSpec.groovy:1927)
[...]

PR #291, opened 2022-10-28
Add test for Micronaut Core regression

graemerocher:


graemerocher:
See conversation here https://github.com/micronaut-projects/micronaut-serialization/pull/290

Issue #299, opened 2022-11-01
JsonAnyGetter does not work for for Serde but works for Jackson


sdelamo:
For this controller: 

```java
package com.example;

import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;

import java.util.Collections;

@Controller
public class HomeController {

    @Get
    Token index() {
        return new Token(Collections.singletonMap("roles", Collections.singletonList("ADMIN")));
    }

}
```

and this test: 

```java
package com.example;

import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.HttpStatus;
import io.micronaut.http.client.BlockingHttpClient;
import io.micronaut.http.client.HttpClient;
import io.micronaut.http.client.annotation.Client;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import jakarta.inject.Inject;
import org.junit.jupiter.api.Test;

import java.util.Map;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotNull;

@MicronautTest
class HomeControllerTest {

    @Inject
    @Client("/")
    HttpClient httpClient;

    @Test
    void testSerde() {
        BlockingHttpClient client = httpClient.toBlocking();
        HttpResponse<Map> response = client.exchange(HttpRequest.GET("/"), Map.class);
        assertEquals(HttpStatus.OK, response.getStatus());
        assertNotNull(response.getBody().get().get("roles"));

    }
}

```

## Jackson ✅

For jackson it passes. 

```java
package com.example;

import com.fasterxml.jackson.annotation.JsonAnyGetter;
import com.fasterxml.jackson.annotation.JsonAnySetter;
import io.micronaut.core.annotation.Introspected;
import io.micronaut.core.annotation.NonNull;
import io.micronaut.core.annotation.Nullable;

import java.util.HashMap;
import java.util.Map;

@Introspected
public class Token {

    @NonNull
    private final Map<String, Object> extensions = new HashMap<>();

    public Token(@Nullable Map<String, Object> extensions) {
        if (extensions != null) {
            this.extensions.putAll(extensions);
        }
    }

    @JsonAnyGetter
    public Map<String, Object> getExtensions() {
        return extensions;
    }
}
```

## Serde ❌

For Serde we use `@Serdeable` instead of `@Introspected` but it fails: 

```java
package com.example;

[...]

PR #300, opened 2022-11-01
Fix for 299

sdelamo:
see: https://github.com/micronaut-projects/micronaut-serialization/issues/299


graemerocher (changes requested):
can you modify this PR to not need the http server? It seems unnecessary to include it

sdelamo:
> can you modify this PR to not need the http server? It seems unnecessary to include it

I have removed the server.  Now: 

- ✅ Pojo with @Serdable and serde
- ✅ Pojo without annotations and jackson
- ❌ Pojo with @Introspected and jackson

PR #323, opened 2022-11-24
Deserialization should work both for timestamp and string by default

guillermocalvo:


serde-support/src/main/java/io/micronaut/serde/support/serdes/DefaultFormattedTemporalSerde.java
```diff
@@ -125,5 +92,23 @@ public final T deserializeNonNull(Decoder decoder, DecoderContext decoderContext
...
+    @NonNull
+    private DateTimeFormatter getFormatter(@NonNull SerdeConfiguration configuration) {
+        // Creates a custom formatter or returns the default one
+        return configuration.getDateFormat()
+            .map(format -> this.createFormatter(format, configuration))
+            .orElseGet(this::getDefaultFormatter);
+    }
+
+    @NonNull
+    private DateTimeFormatter createFormatter(@NonNull String pattern, @NonNull SerdeConfiguration configuration) {
```
sdelamo, line 105: would not be easier to have this method be 

```@NonNull private Optional<DateTimeFormatter> createFormatter( @NonNull SerdeConfiguration configuration) {```

and replace 

```java
@NonNull
     private DateTimeFormatter getFormatter(@NonNull SerdeConfiguration configuration) {
         // Creates a custom formatter or returns the default one. 
         return configuration.getDateFormat()
             .map(format -> this.createFormatter(format, configuration))
             .orElseGet(this::getDefaultFormatter);
     }
```

with : 

```java
/**
 * Creates a custom formatter or returns the default one.
 */
@NonNull
     private DateTimeFormatter getFormatter(@NonNull SerdeConfiguration configuration) {
         return createFormatter(configuration)
             .orElseGet(this::getDefaultFormatter);
     }
```

sdelamo (changes requested):

serde-support/src/main/java/io/micronaut/serde/support/serdes/DefaultFormattedTemporalSerde.java
guillermocalvo: Thanks for your feedback. I've just pushed some changes to refactor these methods.

sdelamo (approved):

graemerocher:
there are compilation errors

sdelamo:
> there are compilation errors

They are fixed.

yawkat:
@guillermocalvo why was this change made? it breaks in #425

yawkat:
the reason why #425 is failing is that `catch (InvalidFormatException e)` and then retrying is not supported by the Decoder interface, and does not work anymore in that pr. i'll take a look at what exactly jackson supports, thanks.

yawkat:
@guillermocalvo the behavior doesn't actually match jackson though. jackson uses nanosecond deserialization by default, while this pr uses milliseconds. do you have a case where this is used in interaction with jackson-databind?

guillermocalvo:
> this pr uses milliseconds

@yawkat Not sure what you mean by that 🤔 `DefaultFormattedTemporalSerde` provides an abstract method `deserializeNonNullWithoutFormat` to delegate actual deserialization to subclasses. They're free to use milliseconds, nanoseconds, or anything in between, as far as the super class is concerned.

In any case, that was existing behavior and out of scope. The only point of this PR was to allow for both number/text representations to be deserialized out of the box, making no assumptions as to how subclasses should be doing it.

PR #324, opened 2022-11-24
Import additional Serdes for CloudEvents issued by Google

guillermocalvo:


sdelamo:
In [the past we decided AWS Lambda Events `@SerdeImport` code belonged in the AWS Module](https://github.com/micronaut-projects/micronaut-serialization/pull/76#issuecomment-1017560244). For consistency, I think we should keep this in the GCP module or have the AWS Lambda Events in the Micronaut Serialization as well. I lean towards keeping them in the cloud modules. 

thoughts @graemerocher ?

sdelamo (approved):

serde-gcp-cloudevents/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+plugins {
```
sdelamo, line 1: This whole file could be replaced with: 


```kotlin
plugins {
    id("io.micronaut.build.internal.module")
}
dependencies {
    annotationProcessor(projects.serdeProcessor)
    implementation(projects.serdeJackson)
    api(libs.managed.google.cloudevent.types)
}
micronautBuild {
    binaryCompatibility {
        enabled.set(false)
    }
}
//TODO remove with Micronaut 4
configurations.named("testRuntimeClasspath") {
    resolutionStrategy.dependencySubstitution {
        substitute(module("io.micronaut:micronaut-jackson-databind"))
            .using(project(":serde-jackson")).because("we want to test Micronaut without jackson databind")
    }
}

```

sdelamo (changes requested):

sdelamo:
I have changed the pr base to 1.5.x which will be the next minor of this module.

sdelamo (approved):

graemerocher:
that is better, though I am unsure if this belongs here our in the GCP module

sdelamo:
> that is better, though I am unsure if this belongs here our in the GCP module

I think this belongs in the GCP module.

sdelamo:
@guillermocalvo I talked to @graemerocher

We are going to keep all the code in the GCP repository but in two modules.

`gcp-serde-cloudevents`
`gcp-function-cloudevents`

PR #337, opened 2022-12-06
Fix support for bootstrap context

graemerocher:
Fixes the error reported here https://ge.micronaut.io/s/pefm2573merk4/tests/:test-resources-hashicorp-vault:test/io.micronaut.testresources.hashicorp.vault.VaultStartedTest/initializationError?top-execution=1

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,11 +3,14 @@ micronaut = "4.0.0-SNAPSHOT"
 micronaut-docs = "2.0.0"
 micronaut-gradle-plugin = "3.6.5"
 micronaut-test = "3.7.0"
-
+micronaut-discovery = "4.0.0-SNAPSHOT"
 groovy = "4.0.5"
 spock = "2.2-groovy-4.0"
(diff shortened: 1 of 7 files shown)


brandon-jorgenson:
@graemerocher I am running into this issue when using serde and the discovery client for hashicorp vault. It looks like this commit never made it into a build. The `SerdeConfiguration` interface is missing the `@BootstrapContextCompatible` that this commit added. I am on release 1.5.2 and I'm getting this error:

`io.micronaut.runtime.Micronaut","message":"Error starting Micronaut server: Failed to inject value for parameter [configuration] of class: io.micronaut.serde.support.DefaultSerdeIntrospections\r\n\r\nMessage: No bean of type [io.micronaut.serde.config.SerdeConfiguration] exists. Make sure the bean is not disabled by bean requirements (enable trace logging for 'io.micronaut.context.condition' to check) and if the bean is enabled then ensure the class is declared a bean and annotation processing is enabled (for Java and Kotlin the 'micronaut-inject-java' dependency should be configured as an annotation processor).`

PR #364, opened 2023-02-10
Remove dependency substitutions from Micronaut Framework 4.0 documentation

### LINKED ISSUE #361: remove unnecessary  dependency substitution hints from Micronaut Framework 4.0 compatible documentation
### Issue description

In Micronaut Framework 4 the following will not be necessary: 

```groovy
configurations.all {
    resolutionStrategy.dependencySubstitution {
        substitute(module("io.micronaut:micronaut-jackson-databind"))
            .using(module("io.micronaut.serde:micronaut-serde-jackson:{version}"))
    }
}
```

We should remove those snippets from the docs

wetted:
closes #361

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -6,7 +6,7 @@ pluginManagement {
 }
 
 plugins {
-    id("io.micronaut.build.shared.settings") version "6.1.1"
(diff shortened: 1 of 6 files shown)


wetted:
I'm not sure what broke that test. All I changed was docs, and the version of build-plugins (which didn't break it either).

PR #368, opened 2023-02-13
support enums with @JsonCreator

### LINKED ISSUE #360: @JsonCreator being ignored on enums
### Expected Behavior

Using `@JsonCreator` Annotation to build Objects.

### Actual Behaviour

`@JsonCreator` is being ignored when it's a static factory. It's always using the Constructor.

Exception:

```
Caused by: java.lang.IllegalArgumentException: No enum constant com.foo.bar.Foo.bar
	at java.base/java.lang.Enum.valueOf(Enum.java:273)
	at io.micronaut.serde.support.serdes.EnumSerde.deserializeNonNull(EnumSerde.java:42)
	at io.micronaut.serde.support.serdes.EnumSerde.deserializeNonNull(EnumSerde.java:36)
	at io.micronaut.serde.util.NullableDeserializer.deserialize(NullableDeserializer.java:38)

```

### Steps To Reproduce

Use the following Enum:

```
@Serdeable
public enum Foo {
  BAR("bar"),
  BAZ("baz");

  private final String value;

  public Foo(String value) {
    this.value = value;
  }

  @JsonCreator
  public static Foo fromValue(String value) {
    return Arrays.stream(values())
        .filter(foo -> Objects.equals(foo.toString(), value))
        .findFirst();
  }
  
  public String toString() {
    return this.value;
  }

}
```

### Environment Information

- Operating System: MacOS
- Java 17

### Example Application

_No response_

### Version

3.8.5

graemerocher:
Fixes #360

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/serdes/EnumSerde.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/serdes/EnumSerde.java
@@ -24,7 +24,6 @@ import java.util.Locale;
 import io.micronaut.core.beans.BeanIntrospection;
 import io.micronaut.core.beans.BeanMethod;
 import io.micronaut.core.beans.exceptions.IntrospectionException;
-import io.micronaut.core.reflect.exception.InstantiationException;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.Executable;
 import io.micronaut.core.util.ArrayUtils;
--- a/serde-support/src/main/java/io/micronaut/serde/support/serdes/EnumSerde.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/serdes/EnumSerde.java
@@ -24,7 +24,6 @@ import java.util.Locale;
 import io.micronaut.core.beans.BeanIntrospection;
 import io.micronaut.core.beans.BeanMethod;
 import io.micronaut.core.beans.exceptions.IntrospectionException;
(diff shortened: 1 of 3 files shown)


serde-support/src/main/java/io/micronaut/serde/support/serdes/EnumSerde.java
```diff
@@ -67,6 +68,30 @@ public E deserializeNonNull(Decoder decoder, DecoderContext decoderContext, Argu
         }
     }
 
+    @Override
+    public Deserializer<E> createSpecific(DecoderContext context, Argument<? super E> type) {
+        try {
+            BeanIntrospection<? super E> si = introspections.getDeserializableIntrospection(type);
+            return (decoder, context1, type1) -> {
+                String s = decoder.decodeString();
```
yawkat, line 76: what if it's not String

yawkat: yes with valueOf only String made sense, but if we support JsonCreator, there should also be support for non-String inputs to that creator

graemerocher: Ok implemented support for other types, but not sure how to implement support for multiple arguments so restricted to work argument

yawkat (approved):

Issue #371, opened 2023-02-15
BsonReaderDecoder throws BsonInvalidOperationException when skipping an object or array value


rohitkumbhar:
### Expected Behavior

`BsonReaderDecoder ` should be able to skip any non-primitive value

### Actual Behaviour

`BsonReaderDecoder` gets an exception when trying to skip an object or array value

```
readStartDocument can only be called when State is VALUE, not when State is END_OF_DOCUMENT.
org.bson.BsonInvalidOperationException: readStartDocument can only be called when State is VALUE, not when State is END_OF_DOCUMENT.
	at app//org.bson.AbstractBsonReader.throwInvalidState(AbstractBsonReader.java:668)
	at app//org.bson.AbstractBsonReader.verifyBSONType(AbstractBsonReader.java:686)
	at app//org.bson.AbstractBsonReader.checkPreconditions(AbstractBsonReader.java:721)
	at app//org.bson.AbstractBsonReader.readStartDocument(AbstractBsonReader.java:449)
	at app//io.micronaut.serde.bson.BsonReaderDecoder.nextToken(BsonReaderDecoder.java:99)
	at app//io.micronaut.serde.support.AbstractStreamDecoder.skipValue(AbstractStreamDecoder.java:844)
	at io.micronaut.serde.bson.BsonReaderDecoderJsonSpec.skip unknown object decode(BsonReaderDecoderJsonSpec.groovy:150)
``` 

### Steps To Reproduce

Failing test:

```groovy
 def 'skip unknown object decode'() {
        given:
        // data has a key "nested" with an object as a value 
        def data = '{ "name": "Foo", "nested": {"key": "value"}}'
        when:
        def w = createDecoder(data).decodeObject()
        then:

        while (true) {
            def key = w.decodeKey()
            if (key == "nested") {
                // Attempt to skip nested
                w.skipValue()
                break
            } else {
                w.decodeString()
            }
        }
    }
```

The `skipValue()`  call causes the exception above. 
NOTE: If I change the value type of the `nested` key to a `String`, the `skipValue()` call does not throw an exception

### Version

micronaut-serde-bson 1.3.3

PR #372, opened 2023-02-15
Issue #371:  Sets the currentToken to null after skipping a value

rohitkumbhar:
The value of `currentToken` is used to read the next token. The MongoDB BSON reader consumes the full value and sets the internal state to TYPE. This change sets the `currentToken` to `null` so the next token can be read correctly


graemerocher:
Thanks for the contribution!

PR #374, opened 2023-02-17
Add property filter

andriy-dmytruk:
Adds support for the [JsonFilter](https://fasterxml.github.io/jackson-annotations/javadoc/2.7/com/fasterxml/jackson/annotation/JsonFilter.html) annotation on types.

* The annotation seems to be only used during serialization, therefore deserialization is not modified.
* The annotation is allowed on properties but this wasn't implemented.
* [Jackson PropertyFilter](https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/ser/PropertyFilter.html) can be used to filter out container elements. It wasn't implemented yet.

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/PropertyFilter.java
+++ b/serde-api/src/main/java/io/micronaut/serde/PropertyFilter.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


serde-api/src/main/java/io/micronaut/serde/PropertyFilter.java
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @param encoderContext the encoder context
+     * @param value the object being serialized
+     * @param propertySerializer the serializer of the property
+     * @param propertyValue the property being serialized
+     * @param propertyName the name of the property
+     *
+     * @return whether the property should be included in serialization
+     */
+    boolean doIncludeProperty(Serializer.EncoderContext encoderContext, Object value, Serializer<Object> propertySerializer, Object propertyValue, String propertyName);
```
graemerocher, line 39: refactor to:

```java

    /**
     *
     * @param encoderContext the encoder context
     * @param propertySerializer the serializer of the property
     * @param bean the object being serialized
     * @param propertyName the name of the property
     * @param propertyValue the property being serialized
     *
     * @return whether the property should be included in serialization
     */
    boolean shouldInclude(@NonNull Serializer.EncoderContext encoderContext, @NonNull Serializer<Object> propertySerializer, @NonNull Object bean , @NonNull String propertyName, @Nullable Object propertyValue);
```

PR #384, opened 2023-02-24
Fix Property Filter usage with json property renaming

andriy-dmytruk:
Given the following property:
```java
@JsonProperty("my-property")
String myProperty;
```
the `PropertyFilter` should get `myProperty` as name instead of `my-property`.


yawkat:
is this the jackson behavior? maybe the filter should just have access to both names.

andriy-dmytruk:
On the second check, the previous behavior is correct. Sorry for the mistake 🤦

In jackson you get access to the value of `@JsonProperty` through `BeanPropertyWriter.getName` and not the actual property name. We could provide the `Argument` to the user as well though, which would mean they could access both names.

PR #395, opened 2023-03-06
JSON value and builder annotations support for enums for 1.5.x

andriy-dmytruk:
- Backport #368.
- Add support for `null` values in the json creator and json value with enums.
- Extend the tests.


andriy-dmytruk:
Added changes so that json builder and creator could support `null` values.

yawkat (approved):

PR #397, opened 2023-03-10
Oracle json serde

radovanradic:
During testing Oracle JSON parsing I have noticed our parser did not support some date types, binary etc.
Most of default serde for date expect long but from Oracle JSON we are getting date formatted in string so I had to support parsing dates using `coerceScalarToString` in `OracleJdbcJsonParserDecoder` and then registering ser/deser for given date types. To use that, users would have to declare these serializers on fields since these have low ordering as we don't want to change current behavior. This can be explained in docs in micronaut-data later, if this approach makes sense. I couldn't figure out better approach at the moment.
This is also designed to serialize back into the DATA string ready to be updated in the database.


serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/OracleJdbcJsonParserDecoder.java
```diff
@@ -102,13 +103,27 @@ protected String coerceScalarToString() {
...
+                return jsonParser.getLocalDateTime().toString();
             case VALUE_TRUE:
                 return StringUtils.TRUE;
             case VALUE_FALSE:
                 return StringUtils.FALSE;
-            default:
+            case VALUE_BINARY:
+                // This is used to parse metadata _etag from Oracle Json View which is treated as binary value
+                // and getString() returns what we need.
+                OracleJsonBinary binary = jsonParser.getValue().asJsonBinary();
```
radovanradic, line 124: This does not return consistent results from Oracle JSON. For example, etag field is binary type and it seems is is already encoded using base16. This code works ok for etag but for memo, blob etc it returns base16 encoded strings. If we want to decode, I would need to use google guava library and don't know if that's allowed.

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Serdeable.Deserializable(using = OracleJsonOffsetDateTimeSerde.class)
+    @Serdeable.Serializable(using = OracleJsonOffsetDateTimeSerde.class)
+    private OffsetDateTime offsetDateTime;
+
+    @Serdeable.Deserializable(using = OracleJsonDateSerde.class)
+    @Serdeable.Serializable(using = OracleJsonDateSerde.class)
+    private Date date;
+
+    @Serdeable.Deserializable(using = OracleJsonInstantSerde.class)
```
radovanradic, line 35: This is how we need to configure fields if we want to read from Oracle JSON and be able to write back to. It means these beans might require DTO for data transfer over web because it uses serialization for DATA json.

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/OracleJdbcJsonParserDecoder.java
dstepanov: Is it specific to the custom value? Maybe we can check for the property name?

radovanradic: I have exposed `decodeBinary` method in the decoder and now it can be done easily and more efficient (no need to convert to string etc.)

radovanradic:
Closing, messed up something when pulling latest from master. Will create a new PR later.

dstepanov:
You can always use `force push` instead of creating a new PR

PR #398, opened 2023-03-13
Support for missing Oracle JSON types serde

radovanradic:
During testing Oracle JSON parsing I have noticed our parser did not support some date types, binary etc.
Most of default serde for date expect long but from Oracle JSON we are getting date formatted in string so I had to support parsing dates using coerceScalarToString in OracleJdbcJsonParserDecoder and then registering ser/deser for given date types. To use that, users would have to declare these serializers on fields since these have low ordering as we don't want to change current behavior. This can be explained in docs in micronaut-data later, if this approach makes sense. I couldn't figure out better approach at the moment.
This is also designed to serialize back into the DATA string ready to be updated in the database.


serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Serdeable.Deserializable(using = OracleJsonOffsetDateTimeSerde.class)
+    @Serdeable.Serializable(using = OracleJsonOffsetDateTimeSerde.class)
+    private OffsetDateTime offsetDateTime;
+
+    @Serdeable.Deserializable(using = OracleJsonLocaleDateSerde.class)
+    @Serdeable.Serializable(using = OracleJsonLocaleDateSerde.class)
+    private LocalDate date;
+
+    @Serdeable.Deserializable(using = OracleJsonInstantSerde.class)
```
radovanradic, line 36: This is how we need to configure fields if we want to read from Oracle JSON and be able to write back to. It means these beans might require DTO for data transfer over web because it uses serialization for DATA json.

radovanradic:
This is failing now

JsonIgnoreSpec test @JsonIgnore without @Inherited on interface method is not inherited FAILED

and I suspect it is related to some recent changes in core. Should not be related to this PR.

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
graemerocher: this shouldn't be necessary if we order the serdes and perhaps add a way for serde to only be active if the encoder/decoder is of a certain type

graemerocher:
> This is failing now
> 
> `JsonIgnoreSpec test @JsonIgnore without @Inherited on interface method is not inherited FAILED`
> 
> and I suspect it is related to some recent changes in core. Should not be related to this PR.

will look into this

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
radovanradic: Yes, this would not be very user friendly, but for entity that will be mapped to (Oracle) JSON we must (de)serialize field using this serde. However, for HTTP client, we would use different one and separate DTO so we might need to do it this way and document with examples.

dstepanov:
> This is failing now
> 
> `JsonIgnoreSpec test @JsonIgnore without @Inherited on interface method is not inherited FAILED`
> 
> and I suspect it is related to some recent changes in core. Should not be related to this PR.

Please modify this test to pass, I did return the previous functionality.

radovanradic:
> > This is failing now
> > `JsonIgnoreSpec test @JsonIgnore without @Inherited on interface method is not inherited FAILED`
> > and I suspect it is related to some recent changes in core. Should not be related to this PR.
> 
> Please modify this test to pass, I did return the previous functionality.

Changed, thanks.

dstepanov:
Can we somehow implement that serdes logic in the mapper?

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonBinarySerde.java
```diff
@@ -0,0 +1,54 @@
...
+                                          @NonNull Argument<? super byte[]> type) {
+        return decoder.decodeBinary();
+    }
+
+    @Override
+    protected void doSerializeNonNull(@NonNull Encoder encoder, @NonNull EncoderContext context,
+                                      @NonNull Argument<? extends byte[]> type, @NonNull byte[] value) throws IOException {
+        // Expects to be base16 encoded when writing to the db
+        String strValue = OracleJsonBinaryImpl.getString(value, false);
+        encoder.encodeString(strValue);
```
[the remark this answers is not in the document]
radovanradic: Yes, as I have looked at Json View docs, this is how update would work
```
UPDATE USR_VIEW uv SET uv.DATA = '{"_metadata":{"etag":"ECB4301EEDDE6AF0DDB20B20F844ACC4","asof":"000000000020F008"},"usrId":2,"name":"New User3","ym":"P1Y2M","ds":"PT24H","bd":9.9999,"memo":"6D656D6F","ts":"2023-03-12T21:59:55.121934","date":"2023-03-12","tstz":"2023-03-12T21:59:55.505109+01:00"}'
WHERE uv.DATA.id = 2
```
And I tested it locally. For examples it succeeds when etag matches and throws an error when I changed it to something unexpected.

radovanradic:
> Please rename it to `is inherited`

Renamed.

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonInstantSerde.java
```diff
@@ -0,0 +1,55 @@
...
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class OracleJsonInstantSerde extends OracleJsonTemporalSerde<Instant> {
+
+    @Override
+    @NonNull
+    protected Instant doDeserializeNonNull(@NonNull OracleJdbcJsonParserDecoder decoder,
+                                           @NonNull DecoderContext decoderContext,
+                                           @NonNull Argument<? super Instant> type) {
+        LocalDateTime localDateTime = (LocalDateTime) decoder.decodeTemporal();
+        return localDateTime.atZone(ZoneId.systemDefault()).toInstant();
```
yawkat, line 48: i don't think we should use the system time zone ever

radovanradic:
> Can we somehow implement that serdes logic in the mapper?

Would be good if we can force somehow these serdes for Oracle JSON, but not sure how can be done because entities will be deserialized using SimpleObjectDeserializer and it goes through props to get deserializer for each.

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonInstantSerde.java
yawkat: in fact this call can fail if the LDT is exactly during an hour that is skipped by a DST change

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
yawkat: i think it should be opt-in like in these annotations, since not everyone may want to use the db-specific data types, and because they might lose information (e.g. precision or some tz info)

graemerocher:
I am working on some changes to this PR

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonInstantSerde.java
yawkat: or to put it more strongly, imo we should not allow conversion from LDT to instant out of the box at all, since it always requires choosing a time zone, which is an arbitrary choice. let the user make that choice by declaring the field as LDT and converting it themselves.

radovanradic: Removed instant serde

yawkat (approved):

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonLocaleDateTimeSerde.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Serde for {@link LocalDateTime} from Oracle JSON.
+ *
+ * @author radovanradic
+ * @since 2.0.0
+ */
+@Singleton
+@Order(-100)
+public class OracleJsonLocaleDateTimeSerde extends OracleJsonTemporalSerde<LocalDateTime> {
```
radovanradic, line 33: Maybe these serdes could be internal classes because using new `OracleType` annotation does the job now?

graemerocher: Perhaps, @dstepanov recommends we change the tests and make the `OracleType` annotation unnecessary

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
radovanradic: With some back and forth, this is not needed anymore as Graeme designed serde registration that will allow us to use serdes within Oracle JSON parsing context without need to register.

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonLocaleDateTimeSerde.java
radovanradic: Changed in https://github.com/micronaut-projects/micronaut-serialization/pull/398/commits/905d53561b485d82cfa42f59503e25eedf4bfc73

serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/OracleJdbcJsonParserDecoder.java
```diff
@@ -167,11 +163,127 @@ protected void skipChildren() {
...
+                    bytes[i] = 0;
+                } else {
+                    bytes[i] = (byte) oracleJsonArray.getInt(i);
+                }
+            }
+            nextToken();
+            return bytes;
+        }
+        if (currentEvent == OracleJsonParser.Event.VALUE_STRING) {
+            String str = jsonParser.getString();
```
radovanradic, line 201: I wanted to avoid this, but for tests that parse serialized JSON with byte[] we need to have this because byte[] needs to be base16 encoded when writing to the db so for the test for all decoders we need to be able to decode it back and verify.

dstepanov (approved):

PR #400, opened 2023-03-14
Remove need for oracle type for Oracle JSON serde

radovanradic:
An attempt to remove need for specific configuration of entities that are being de/serialized from Oracle JSON.


serde-oracle-jdbc-json/src/main/java/io/micronaut/serde/oracle/jdbc/json/serde/OracleJsonBinarySerde.java
```diff
@@ -41,31 +38,16 @@ public class OracleJsonBinarySerde extends AbstractOracleJsonSerde<byte[]> {
...
-                encoder,
-                context,
-                type,
-                value
-            );
-        }
-
+                                      @NonNull Argument<? extends byte[]> type, @NonNull byte[] value) {
+        // Expects to be base16 encoded when writing to the db
+        String strValue = OracleJsonBinaryImpl.getString(value, false);
```
radovanradic, line 49: We reuse logic from AbstractOracleJsonSerde and if oracle decoder not present then we can use default one, and this is in each oracle serde implementations.

serde-oracle-jdbc-json/src/test/java/io/micronaut/serde/bson/SampleData.java
```diff
@@ -9,28 +9,20 @@
...
 
 import io.micronaut.core.annotation.Introspected;
-import io.micronaut.serde.oracle.jdbc.json.annotation.OracleType;
-
-import static io.micronaut.serde.oracle.jdbc.json.annotation.OracleType.Type.*;
 
 @Introspected
 public class SampleData {
 
-    @OracleType(TEMPORAL)
```
radovanradic, line 19: So this is not needed anymore. However, these beans/entities won't work in HTTP client deserialization since it would trigger Oracle serde. So, advice would be to use DTO for transfer over HTTP. Think we don't have other way because Oracle serdes will be in the context.

PR #408, opened 2023-03-20
build: Micronaut Framework 4.0.0-M1

sdelamo:
- GitHub Actions sync

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,18 +161,18 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/')
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       - name: Download artifacts
-        uses: actions/download-artifact@v3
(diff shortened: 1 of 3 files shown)


sdelamo:
@melix this is attempting to resolve platform with micronaut version  4.0.0-M1 

 > Could not find io.micronaut.platform:micronaut-platform:4.0.0-M1.

What should we do? We could workaround not including docs-examples modules and releasing the module.

melix:
I have pushed a workaround. Basically we have an inconsistency now, because our build plugins assume that the version of `micronaut` is for `micronaut-core`, but the Gradle plugins (righteously) assume that it's the _platform_ version. The fact that it fails simply illustrates a problem with using the Gradle plugin for testing: what we use then is something which is designed for _releases_, not development: it makes use of the platform, which is the ultimate deliverable of a Micronaut release.

The workaround is simply to use a SNAPSHOT version of Micronaut in the tests, until we have a milestone release of the platform. However, there will most likely always be a mistmatch between the version of Micronaut that a module builds against (here M1) and the version that we can test against using the Gradle plugin (which is a platform, so released with _older_ modules).

sdelamo:
> what we use then is something which is designed for _releases_, not development:

Yes, I agree. We should not use [Micronaut Gradle Plugin](https://plugins.gradle.org/plugin/io.micronaut.application) in the projects. We only use them in a subset of repositories.

PR #425, opened 2023-04-20
Optimizations

yawkat:
This PR has some simple optimizations in the first commits. However to get to jackson-databind level, it needs the changes in the last commit. The last commit rewrites JacksonDecoder to replace the nextToken/getValue pattern with calls to the equivalent nextValue methods, which are faster. There are two problems with this change:

- it's now invalid to call another decodeX operation after a failed decode call. This was my intention with the Decoder api from the beginning, you should not "speculatively" call the decode methods and retry if there is a failure. However DefaultFormattedTemporalSerde depends on this behavior unfortunately. This is the test failure. I think DefaultFormattedTemporalSerde should be changed, but it looks incompatible.
- we need to avoid decodeNull calls to take advantage of this change. I added decodeStringNullable and similar methods to Decoder, and changed some serdes to allowNull and use these new methods. This change is about halfway done, finishing it would probably give additional perf benefit.

WDYT?


yawkat:
maybe instead of changing the serdes to all allowNull=true, there could be a new method decodeNullable that is essentially equivalent to `return allowNull() ? decode(...) : decoder.decodeNull() ? null : decode(...)`, and that would be overridden by the different serdes

graemerocher (approved):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/PropertiesBag.java
```diff
@@ -52,6 +61,44 @@ private PropertiesBag(BeanIntrospection<T> beanIntrospection,
...
+        if (properties.length > 0 && properties.length <= 64) {
+            this.propertiesMask = -1L >>> (64 - properties.length);
+        } else {
+            this.propertiesMask = 0;
+        }
+        Stream<String> propStream = beanIntrospection.getBeanProperties().stream().map(Named::getName);
+        if (nameToPropertiesMapping != null) {
+            propStream = Stream.concat(propStream, nameToPropertiesMapping.keySet().stream());
+        }
+        Set<String> props = propStream.collect(Collectors.toSet());
```
dstepanov, line 73: This is not needed anymore after the changes in Core

dstepanov (changes requested):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/PropertiesBag.java
yawkat: i think it should stay, because propertyIndexOfSlow still does a bit more work than just calling the core propertyIndexOf. it could use the new StringIntMap, but that one isn't available yet.

dstepanov: ok

dstepanov (approved):

PR #449, opened 2023-05-12
Fixed concurrency issue in ObjectDeserializer

msupic:


graemerocher (approved):

graemerocher:
stack overflows so probably not the solution

dstepanov:
I will try to fix it

yawkat:
@timyates no, there is still a data race between the put and the unguarded get in that fix. put must come after the initialize somehow

dstepanov:
@graemerocher Fixed

PR #456, opened 2023-05-15
Always use one `DeserBean` instance

dstepanov:
I think we need to have one instance per class otherwise in a case of a cycle we would create infinite amount of `DeserBean`s


dstepanov:
Corrected

PR #496, opened 2023-06-19
Introduce nesting depth limits for encoders and decoders

yawkat:
Create a new base class LimitingStream for all Decoder and Encoder implementations that checks the nesting depth. The nesting depth is configured using a new configuration property.

The purpose of this is to harden against issues like https://github.com/FasterXML/jackson-databind/issues/3972 . Another benefit is that accidental recursive data structures made by developers won't lead to a stack overflow but will be caught earlier with a nicer error message.

This patch changes many classes, mostly because the new limit has to be propagated from the serde configuration to the encoders/decoders. To simplify addition of other limits in the future, I've wrapped the depth limit into an opaque data structure. Other possible limits would be e.g. on total output size or on array sizes.

I've tried to keep this patch as compatible as possible even though it's only going into a major version, to make the release process easier. Methods that need the new limit parameter and were not internal I've deprecated.

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/LimitingStream.java
+++ b/serde-api/src/main/java/io/micronaut/serde/LimitingStream.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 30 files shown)


graemerocher:
what is the impact on performance?

yawkat:
@graemerocher no difference in the decodeUsers benchmark

PR #524, opened 2023-07-19
Support decoding base64 to byte[]

### LINKED ISSUE #520: Deserialization of Base 64-encoded strings into byte arrays is not working
Micronaut Serialization fails when trying to deserialize a String that is Base 64-encoded into a `byte[]`, such as:

`eyJzaXplIjoxMDAwLCJlbmNyeXB0ZWQiOmZhbHNlLCJ3b3JrZmxvd0lkIjoxfQ==` which decodes as `{"size":1000,"encrypted":false,"workflowId":1}`

yawkat:
jackson-databind writes byte[] as base64 and supports reading from base64 or array. Serde 2 writes byte[] as array, but does not support reading from base64, so prior to this change we are only compatible with jackson in one direction (serde encoder -> jackson decoder).

This patch adds decode support for base64, and for the bson binary type. For oracle jdbc json, a decodeBinary method already exists.

On the encode side, there is now a config flag to use format-specific byte array encoding, which is base64 for json. The default remains to write an array of numbers.

Fixes #520

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/Decoder.java
+++ b/serde-api/src/main/java/io/micronaut/serde/Decoder.java
@@ -19,6 +19,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.type.Argument;
 import io.micronaut.json.tree.JsonNode;
+import io.micronaut.serde.util.BinaryCodecUtil;
 
 import java.io.IOException;
 import java.math.BigDecimal;
@@ -281,6 +282,38 @@ public interface Decoder extends AutoCloseable {
     @Nullable
     default BigDecimal decodeBigDecimalNullable() throws IOException {
         return decodeNull() ? null : decodeBigDecimal();
+    }
+
+    /**
+     * Decode binary data from this stream. Binary data can be serialized in multiple different
+     * ways that differ by format.
+     *
+     * <ul>
+     *     <li>An array of numbers must be supported by all implementations, for compatibility.
+     *     This is also the default implementation.</li>
+     *     <li>A base64 string. This is convenient for text-based formats like json, and is
+     *     supported by jackson.</li>
+     *     <li>A format-specific type, for binary formats such as bson.</li>
+     *     <li>Other format specific behavior. Oracle JDBC Json will parse strings as hex, for
+     *     example.</li>
+     * </ul>
+     *
+     * Implementations <b>must</b> support the array shape, but the other shapes are optional.
+     *
+     * @return The decoded byte array
+     */
+    default byte @NonNull [] decodeBinary() throws IOException {
+        return BinaryCodecUtil.decodeFromArray(this);
+    }
+
+    /**
+     * Equivalent to {@code decodeNull() ? null : decodeBinary()}.
+     *
+     * @return The value
+     * @throws IOException If an unrecoverable error occurs
+     */
+    default byte @Nullable [] decodeBinaryNullable() throws IOException {
+        return decodeNull() ? null : decodeBinary();
     }
 
     /**
--- a/serde-api/src/main/java/io/micronaut/serde/Decoder.java
+++ b/serde-api/src/main/java/io/micronaut/serde/Decoder.java
@@ -301,6 +301,7 @@ public interface Decoder extends AutoCloseable {
      * Implementations <b>must</b> support the array shape, but the other shapes are optional.
      *
      * @return The decoded byte array
+     * @since 2.1
      */
     default byte @NonNull [] decodeBinary() throws IOException {
         return BinaryCodecUtil.decodeFromArray(this);
@@ -311,6 +312,7 @@ public interface Decoder extends AutoCloseable {
      *
      * @return The value
      * @throws IOException If an unrecoverable error occurs
+     * @since 2.1
      */
     default byte @Nullable [] decodeBinaryNullable() throws IOException {
         return decodeNull() ? null : decodeBinary();
--- a/serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
+++ b/serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
@@ -60,6 +60,17 @@ public interface SerdeConfiguration {
     NumericTimeUnit getNumericTimeUnit();
 
     /**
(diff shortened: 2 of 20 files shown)


serde-api/src/main/java/io/micronaut/serde/Decoder.java
```diff
@@ -283,6 +284,38 @@ default BigDecimal decodeBigDecimalNullable() throws IOException {
...
+     *     This is also the default implementation.</li>
+     *     <li>A base64 string. This is convenient for text-based formats like json, and is
+     *     supported by jackson.</li>
+     *     <li>A format-specific type, for binary formats such as bson.</li>
+     *     <li>Other format specific behavior. Oracle JDBC Json will parse strings as hex, for
+     *     example.</li>
+     * </ul>
+     *
+     * Implementations <b>must</b> support the array shape, but the other shapes are optional.
+     *
```
alvarosanchez, line 302: Need `@since` tags for new methods

alvarosanchez (approved):

serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
```diff
@@ -59,6 +59,17 @@ public interface SerdeConfiguration {
...
+    /**
+     * Control whether to use legacy behavior for writing byte arrays. When set to {@code true} (the
+     * default in serde 2.x), byte arrays will always be written as arrays of numbers. When set to
+     * {@code false}, the encoding may be format-specific instead, and will be a base64 string for
+     * JSON.
+     *
+     * @return Whether to use legacy byte array writing behavior
+     */
+    @Bindable(defaultValue = "true")
+    boolean isWriteLegacyByteArrays();
```
graemerocher, line 71: Rename to `isWriteByteArraysAsBinary` and flip logic. Makes more logical sense from a configuration perspective and from the API perspective with the encodeBinary/decodeBinary methods.

serde-api/src/main/java/io/micronaut/serde/util/BinaryCodecUtil.java
```diff
@@ -0,0 +1,68 @@
...
+        try (Decoder arrayDecoder = base.decodeArray(BYTE_ARRAY)) {
+            while (arrayDecoder.hasNextArrayValue()) {
+                Byte b = arrayDecoder.decodeByteNullable();
+                buffer.write(b == null ? 0 : b);
+            }
+        }
+        return buffer.toByteArray();
+    }
+
+    public static byte[] decodeFromString(Decoder base) throws IOException {
```
graemerocher, line 48: rename to `decodeFromBase64String` to make it clearer what is happening here

serde-api/src/main/java/io/micronaut/serde/util/BinaryCodecUtil.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public static void encodeToArray(Encoder encoder, byte[] data) throws IOException {
+        try (Encoder arrayEncoder = encoder.encodeArray(BYTE_ARRAY)) {
+            for (byte i : data) {
+                arrayEncoder.encodeByte(i);
+            }
+        }
+    }
+
+    public static void encodeToString(Encoder encoder, byte[] data) throws IOException {
```
graemerocher, line 65: Rename to `encodeToBase64String` to make it clearer what is happening here

serde-api/src/main/java/io/micronaut/serde/util/BinaryCodecUtil.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Base64;
+
+/**
+ * Common implementations for reading/writing byte arrays.
+ */
+@Internal
+public final class BinaryCodecUtil {
```
graemerocher, line 31: does this need to be public?

graemerocher (changes requested):

serde-api/src/main/java/io/micronaut/serde/util/BinaryCodecUtil.java
yawkat: yes it's used in multiple implementations of Encoder/Decoder

serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
yawkat: that's also a bit confusing though, because for json it doesnt write the byte array as binary, it writes it as base64. the encode/decode method is called that because the data is binary, even though the serialized form isnt necessarily

graemerocher: yeah just not keen on this name and including `legacy` in configuration values. It should describe what it does.

yawkat: how about isWriteBinaryAsArray

serde-support/src/test/groovy/io/micronaut/serde/support/CoreTypeSerdeSpec.groovy
```diff
@@ -115,4 +115,16 @@ class CoreTypeSerdeSpec extends Specification {
...
+    void "test byte array deserialization shapes"(byte[] expected, String json) {
+        when:
+        def deser = jsonMapper.readValue(json, Argument.of(byte[]))
+        then:
+        deser == expected
+
+        where:
+        expected | json
+        [0, 1]   | '[0,1]'
+        [0, 1]   | '"' + Base64.encoder.encodeToString([0, 1] as byte[]) + '"'
```
alvarosanchez, line 128:
```suggestion
        [0, 1]   | '"' + Base64.encoder.encodeToString([0, 1] as byte[]) + '"'
        []       | ''        
```

alvarosanchez (changes requested):
It needs to handle empty strings as empty arrays

serde-support/src/test/groovy/io/micronaut/serde/support/CoreTypeSerdeSpec.groovy
yawkat: no it shouldn't

PR #527, opened 2023-07-21
Support for @JsonDeserialize(builder=..)

graemerocher:
Draft PR for builder support, some open questions:

* Don't know how to support pre-instantiation hook
* Unsure about strict nullable

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,5 +5,4 @@ plugins {
 
 repositories {
     mavenCentral()
-    mavenLocal()
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,4 +5,5 @@ plugins {
 
 repositories {
     mavenCentral()
+    mavenLocal()
 }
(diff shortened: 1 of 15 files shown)


build.gradle.kts
```diff
@@ -5,4 +5,5 @@ plugins {
 
 repositories {
     mavenCentral()
+    mavenLocal()
```
yawkat, line 8: remove when this pr is ready

yawkat (approved):

yawkat:
what exactly are the pre-instantiate hooks used for (theyre used in data right)?

i think builders should have the same behavior as creator methods wrt pre-instantiate hooks. i think atm the hook is called after the creator but before any more properties are set.

graemerocher:
I am not sure about these pre-instantiate hooks myself, @dstepanov will have to confirm

dstepanov:
The hook is used to perform possible validation before the constructor is invoked. In this case, we need to think about how we would trigger the builder method validation.

PR #530, opened 2023-07-25
Fix deserialization with nullable constructor argument of type Object

radovanradic:
Fix for issue reported in data https://github.com/micronaut-projects/micronaut-data/issues/2403

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBean.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBean.java
@@ -655,10 +655,11 @@ class DeserBean<T> {
             this.instrospection = instrospection;
             this.index = index;
             this.argument = argument;
-            this.mustSetField = argument.isNonNull() || (!argument.getType().equals(Object.class) && (argument.isAssignableFrom(Optional.class)
-                    || argument.isAssignableFrom(OptionalLong.class)
-                    || argument.isAssignableFrom(OptionalDouble.class)
-                    || argument.isAssignableFrom(OptionalInt.class)));
+            Class<?> type = argument.getType();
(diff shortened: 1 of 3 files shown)


serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBean.java
```diff
@@ -655,10 +655,10 @@ public DerProperty(ConversionService conversionService,
             this.instrospection = instrospection;
             this.index = index;
             this.argument = argument;
-            this.mustSetField = argument.isNonNull() || argument.isAssignableFrom(Optional.class)
+            this.mustSetField = argument.isNonNull() || (!argument.getType().equals(Object.class) && (argument.isAssignableFrom(Optional.class)
```
radovanradic, line 658: This is fix/workaround when field is of type Object like reported in issue https://github.com/micronaut-projects/micronaut-data/issues/2403
Even if it has `@Nullable` annotation and `argument.isNonNull()` is false, second part of disjunction will evaluate to true (because Object is assignable to all of these types) and `mustSetField` will be true.

dstepanov (approved):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBean.java
yawkat: All these Optional types are final and have no other supertypes. Change this code from isAssignableFrom to getType.equals, and you can drop the comparison with Object. it'll be faster, too.

radovanradic: Changed

yawkat (approved):

PR #534, opened 2023-07-27
Issue with @NonNull field in unwrapped class

radovanradic:
@graemerocher Found one issue when field is marked `@NonNull` deserialization throws an error telling that field is required and not set, although it is set.
It seems like some nested objects are being processed more than once during deserializaiton.

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonUnwrappedSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonUnwrappedSpec.groovy
@@ -700,6 +700,15 @@ class InnerFooId {
         read.hashKey.nested.theLong == 200
         read.hashKey.nested.theString == 'MyString'
 
+        cleanup:
+        context.close()
+    }
+
+
+    void "test @JsonUnwrapped - levels 2"() {
+        given:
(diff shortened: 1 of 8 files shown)


graemerocher:
is there supposed to be a failing test here?

radovanradic:
> is there supposed to be a failing test here?

Yes, test `test @JsonUnwrapped - levels` is failing now. Not in github because this is PR against your branch. Basically, this is not needed, you can reproduce in your branch by adding `@NonNull` on any of the fields in nested classes. For example
```
@Serdeable
class InnerFooId {

    @NonNull
    private Long theLong;

    private String theString;

    public Long getTheLong() {
        return theLong;
    }

    public void setTheLong(Long theLong) {
        this.theLong = theLong;
    }

    public String getTheString() {
        return theString;
    }

    public void setTheString(String theString) {
        this.theString = theString;
    }
}
```
and the exception is

> Unable to deserialize type [unwrapped.InnerFooId]. Required property [Long theLong] is not present in supplied data
io.micronaut.serde.exceptions.SerdeException: Unable to deserialize type [unwrapped.InnerFooId]. Required property [Long theLong] is not present in supplied data


because it seems nested objects are processed more than once and in second pass, properties are consumed from the first pass and then it throws error for not finding value for one of the fields.

PR #558, opened 2023-08-31
Support importing packages / builders

graemerocher:
Adds a new experimental `packageName` member to allow importing serialization/deserializers from packages. Also fixes handling of import of builders.


serde-api/src/main/java/io/micronaut/serde/annotation/SerdeImport.java
```diff
@@ -41,7 +42,16 @@
      *
      * @return The type to enable serialization for.
      */
-    Class<?> value();
+    Class<?> value() default void.class;
+
+    /**
+     * Specifies a package name where types can be found. Only publicaly accessible types will be processed.
```
sdelamo, line 48: I assume nested package are not processed. It would be good to specify it in the javadoc.

PR #573, opened 2023-09-15
Test case for private builder constructor method

yawkat:
Unfortunately this test case is a bit removed from my original issue.

In the original issue, I have a third-party bean with a builder and jackson annotations. The builder is generated by lombok and has an inaccessible constructor, I think package-private. I want to import that bean with `@SerdeImport`, with the additional qualification of `@Introspected(builder = @Introspected.IntrospectionBuilder(builderMethod = "builder"))` in order to use the builder method instead of the default constructor of the builder class.

The problem is that serde seems to be ignoring my custom `@Introspected` and replacing it with `@Introspected` with `builderClass` instead of `builderMethod`. In the original project, this leads to a runtime illegal access of the builder constructor.

This test case is modified a bunch, because I wanted to restrict it to one module where the processor is already running. Since the class needs to be annotated with `@JsonDeserialize`, and that triggers serde processing, I can't build this test with `@SerdeImport` anymore. I instead moved the `@Introspected` to the actual bean class. I also made the builder class constructor private, to trigger the error: package-private is not enough since the introspection will be placed in the same package.

In the test case, the bug manifests as a compile-time error:

```
/home/yawkat/dev/mn/micronaut-serialization/doc-examples/example-java/src/main/java/example/BuilderBean.java:12: error: No accessible constructor found for builder: example.BuilderBean$Builder
public class BuilderBean {
```

### DIFF
--- a/doc-examples/example-java/src/main/java/example/BuilderBean.java
+++ b/doc-examples/example-java/src/main/java/example/BuilderBean.java
@@ -1,0 +1,64 @@
+package example;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.micronaut.core.annotation.Introspected;
+
+import java.util.Objects;
+
+@JsonDeserialize(builder = BuilderBean.Builder.class)
+@Introspected(builder = @Introspected.IntrospectionBuilder(builderMethod = "builder"))
+public class BuilderBean {
+    private final String foo;
+
+    private BuilderBean(String foo) {
+        this.foo = foo;
+    }
+
+    public String getFoo() {
+        return foo;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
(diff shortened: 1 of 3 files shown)


yawkat:
ive looked at this a bit further. it seems on the jackson side there is no way to use the `builder()` method–jackson can use a creator sometimes, but the creator has to be in the builder class, not the outer class. 

so we can't use jackson annotations to fix this. it needs a patch that makes the explicit `@Introspected` take precedence over the generated one.

graemerocher:
What happens when you replace https://github.com/micronaut-projects/micronaut-serialization/pull/573/files#diff-b70afbf4ab52166c86b242d23420a7d1e419479e7cf785f2ab8510995c64aae7R8

With just `@Serdeable`

yawkat:
then it works. but i cant modify that annotation, in the original project it's part of an external library. i have to serdeimport the class that has jsondeserialize declared.

PR #594, opened 2023-10-06
Serialization to java 21

timyates:


sdelamo (changes requested):
I think we should have a module which is shared by both kotlin implementations to avoid duplicating tests.

timyates:
> I think we should have a module which is shared by both kotlin implementations to avoid duplicating tests.

We need to process the main source with a different annotation processor (so I believe that would need to be duplicated?)

Do you think it's worth duplicating the main code, but having a Test suite for the tests?

sdelamo (approved):

Issue #616, opened 2023-10-25
bug with Form url encoded and a POJO with a list field


sdelamo:
### Issue description

With this record: 

```java
package com.example;

import io.micronaut.core.annotation.Introspected;
import io.micronaut.serde.annotation.Serdeable;

import java.util.List;

@Serdeable
public record QuestionSave(String question,
                           List<Long> usersId) {
}
```

and this controller: 

```java
package com.example;

import io.micronaut.http.MediaType;
import io.micronaut.http.annotation.Body;
import io.micronaut.http.annotation.Consumes;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Post;

@Controller("/questions")
public class QuestionController {
    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
    @Post("/save")
    QuestionSave save(@Body QuestionSave questionSave) {
        return questionSave;
    }
}
```

and this test fails when the list contains one item: 

```java
package com.example;

import io.micronaut.http.HttpRequest;
import io.micronaut.http.MediaType;
import io.micronaut.http.client.BlockingHttpClient;
import io.micronaut.http.client.HttpClient;
import io.micronaut.http.client.annotation.Client;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import org.junit.jupiter.api.Test;

import java.io.IOException;

import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
import static org.junit.jupiter.api.Assertions.assertEquals;

@MicronautTest
class QuestionControllerTest {

    @Test ❌
    public void formWithListOfOneItem(@Client("/") HttpClient client) throws IOException {
        String body = "question=en+que+trabajas&usersId=1";
        String expectedJson = "{\"question\":\"en que trabajas\",\"usersId\":[1]}";
        assertWithBody(client, body, expectedJson);
    }

    @Test ✅
    public void formWithListOfMoreThanOne(@Client("/") HttpClient client) throws IOException {
        String body = "question=en+que+trabajas&usersId=1&usersId=2";
        String expectedJson = "{\"question\":\"en que trabajas\",\"usersId\":[1,2]}";
        assertWithBody(client, body, expectedJson);
    }

    private static void assertWithBody(HttpClient httpClient, String body, String expectedJson) throws IOException {
        BlockingHttpClient client = httpClient.toBlocking();
[...]

PR #617, opened 2023-10-25
test: run http server tck against Micronaut Serialization Jackson

sdelamo:
see: https://github.com/micronaut-projects/micronaut-serialization/issues/616 and https://github.com/micronaut-projects/micronaut-core/pull/10023

### DIFF
--- a/test-suite-http-server-tck-netty/src/test/java/io/micronaut/serde/http/server/tck/netty/NettyHttpServerTestSuite.java
+++ b/test-suite-http-server-tck-netty/src/test/java/io/micronaut/serde/http/server/tck/netty/NettyHttpServerTestSuite.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.serde.http.server.tck.netty;
+
+import org.junit.platform.suite.api.ExcludeClassNamePatterns;
+import org.junit.platform.suite.api.SelectPackages;
+import org.junit.platform.suite.api.Suite;
+import org.junit.platform.suite.api.SuiteDisplayName;
+
+@Suite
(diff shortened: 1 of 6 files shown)


test-suite-http-server-tck-netty/src/test/java/io/micronaut/serde/http/server/tck/netty/NettyHttpServerTestSuite.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package io.micronaut.serde.http.server.tck.netty;
+
+import org.junit.platform.suite.api.ExcludeClassNamePatterns;
+import org.junit.platform.suite.api.SelectPackages;
+import org.junit.platform.suite.api.Suite;
+import org.junit.platform.suite.api.SuiteDisplayName;
+
+@Suite
+@ExcludeClassNamePatterns("io.micronaut.http.server.tck.tests.constraintshandler.ControllerConstraintHandlerTest")
```
sdelamo, line 24: this test fails for serde https://github.com/micronaut-projects/micronaut-core/blob/4.2.x/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/constraintshandler/ControllerConstraintHandlerTest.java#L118-L119

yawkat: that's a weird test, it relies on a NonNull parameter being set to null and then failing later in constraint validation, which works with jackson-databind (doesnt care about NonNull) but not serde. serde's behavior is fine here.

sdelamo: It isn't a weird test. It describes a scenario many users will anticipate to work. Myself included. You create a controller with a POJO with the nullability annotations. The POJO describes the valid payload of your endpoint. The person invoking the API does not supply a valid JSON. You expect the object to be populated with the available information and the validation submission to be triggered so that you can handle the `ConstraintViolationException`. What are you working on

yawkat: NonNull is not a validation annotation though, it is more fundamental and is checked at a different stage than validation annotations. I see your point, but we can't throw ConstraintViolationException from serde, it's not even on the compile path. And we made the explicit choice to check for null in serde.

Consider it from a different angle. If the JSON included `{\"username\":{},\"password\":\"secret\"}` (entirely wrong *type* for username), then both jackson-databind and serde will fail with not-a-ConstraintViolationException. Serde just extends this same behavior to null (or in this case missing fields). The line between a type error and a constraint error has to be drawn somewhere, and serde (intentionally) checks nulls.

sdelamo:
@yawkat can you a approve or request changes?

PR #620, opened 2023-10-26
Deserialization improvements

dstepanov:
Refactoring/rewriting of the deserialization:
- Detection of the subtype is extracted to a separate deserializer with a newly introduced lookahead deserializer
- `SpecificObjectDeserializer` rewritten to avoid buffering and to simplify the code


yawkat:
i dont like the LookaheadDecoder abstraction. I will try to find an alternative.

dstepanov:
Ok

yawkat:
@dstepanov take a look. This approach avoids new API on Decoder, and it uses the existing decodeBuffer so in theory the jackson implementation can be based on TokenBuffer for performance at some point.

Also do you think the changes to JsonNodeDecoder still have value now that the new subtypes are gone? If not, we can revert them for easier review

dstepanov:
I did a bit of benchmarking of the unwrapped scenario: 10% faster than the previous code.

Vs Jackson:
```
Benchmark                                                (stack)  Mode  Cnt    Score    Error  Units
JacksonBenchmark.decodeUnwrapped     JACKSON_DATABIND_INTROSPECTION  avgt   10  660.406 ± 11.924  ns/op
JacksonBenchmark.decodeUnwrapped                                         SERDE_JACKSON  avgt   10  554.094 ±  7.019  ns/op
```

dstepanov:
@yawkat Looks fine. I think there are some improvements added to `JsonNodeDecoder`, you can copy those subclasses back into the decoder if you want.

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    public static Decoder prime(Decoder decoder) {
+        return new PrimedDecoder(decoder);
+    }
+
+    @Override
+    public @Nullable String decodeKey() throws IOException {
+        DemuxerState.Entry entry;
+        do {
+            entry = state.getEntry(nextKeyIndex++);
```
dstepanov, line 82: @yawkat Can this be rewritten using an iterator?

yawkat: no because the list may change size

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
```diff
@@ -0,0 +1,281 @@
...
+ * @since 2.2.7
+ */
+@Internal
+final class DemuxingObjectDecoder extends DelegatingDecoder {
+    private final DemuxerState state;
+    private int nextKeyIndex;
+
+    private DemuxingObjectDecoder(DemuxerState state, int nextKeyIndex) {
+        this.state = state;
+        this.nextKeyIndex = nextKeyIndex;
```
dstepanov, line 48: @yawkat  incoming `nextKeyIndex` is always zero

dstepanov:
@yawkat IMHO it would be better if `DemuxingObjectDecoder` was split into two implementations, one that caches and one that replays / delegates. Now it's hard to understand with all the states there.

yawkat:
Do you mean just moving the nested classes or actually weakening the API?

dstepanov:
The best would be to have two implementation, the state can be passed as a map of string -> decoder

dstepanov:
Ok let's keep it this way.

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
yawkat: thanks for fixing it

graemerocher (approved):

Issue #263, opened 2022-09-05
`@JsonUnwrapped` cannot serialize subtypes


yawkat:
### Feature description

In cases where serde normally serializes using the dynamic type of a field (e.g. when the class is abstract), `@JsonUnwrapped` should also use the dynamic type.

This is required for https://github.com/micronaut-projects/micronaut-problem-json/pull/196 .

Failing test case:

```groovy
    void "test @JsonUnwrapped - subtyping"() {
        given:
        def context = buildContext("""
package unwrapped;

import com.fasterxml.jackson.annotation.JsonUnwrapped;
import io.micronaut.core.annotation.Introspected;
import io.micronaut.serde.annotation.Serdeable;

@Serdeable
@Introspected(accessKind = Introspected.AccessKind.FIELD)
class Wrapper {
  @JsonUnwrapped
  public final SuperClass name;

  Wrapper(@JsonUnwrapped SuperClass name) {
      this.name = name;
  }
}

@Serdeable
@Introspected(accessKind = Introspected.AccessKind.FIELD)
abstract class SuperClass {
  public final String first;
  SuperClass(String first) {
      this.first = first;
  }
}

@Serdeable
@Introspected(accessKind = Introspected.AccessKind.FIELD)
class SubClass extends SuperClass {
  public final String last;
  SubClass(String first, String last) {
      super(first);
      this.last = last;
  }
}
""")

        when:
        def name = newInstance(context, 'unwrapped.SubClass', "Fred", "Flinstone")
        def wrapper = newInstance(context, 'unwrapped.Wrapper', name)

        def result = writeJson(jsonMapper, wrapper)

        then:
        result == '{"first":"Fred","last":"Flinstone"}'

        cleanup:
        context.close()
    }
```

PR #624, opened 2023-10-30
Support unwrapped subtype beans

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/263

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/ObjectSerializer.java
+++ b/serde-api/src/main/java/io/micronaut/serde/ObjectSerializer.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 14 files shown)


serde-api/src/main/java/io/micronaut/serde/ObjectSerializer.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * A variation of {@link Serializer} that is serializing an object and supports serializing its content into an existing object.
+ *
+ * @param <T> The object type to be serialized
+ * @author Denis Stepanov
+ * @since 2.3
+ */
+@Indexed(ObjectSerializer.class)
+public interface ObjectSerializer<T> extends Serializer<T> {
```
graemerocher, line 32: Perhaps this is a confusing name?

graemerocher (approved):

serde-api/src/main/java/io/micronaut/serde/ObjectSerializer.java
dstepanov: What would you suggest? Before I had `SerializerInto` but I think it makes more sense to call it `ObjectSerializer` because `serializeInto` can only be used with a serializer representing JSON objects `{...}`.

yawkat: UnwrappedObjectSerializer maybe

yawkat (approved):

serde-api/src/main/java/io/micronaut/serde/ObjectSerializer.java
dstepanov: The idea is that an object serializer is any serializer representing an object, and if you put the unwrapped annotation, it will use `serializeInto` method to avoid creating an outer object.

dstepanov: Going to merge it, and rename it later if you have a better suggestion

Issue #49, opened 2021-11-26
Add pretty print/format support


graemerocher:
### Feature description

We should allow outputting formatted JSON

PR #631, opened 2023-11-01
Allow to configure Jackson parser/generator

dstepanov:
Closes https://github.com/micronaut-projects/micronaut-serialization/issues/49

### DIFF
--- a/serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonConfiguration.java
+++ b/serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonConfiguration.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


yawkat (approved):

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Configuration for the Jackson.
+ *
+ * @author Denis Stepanov
+ * @since 2.3
+ */
+@BootstrapContextCompatible
+@Internal
+@ConfigurationProperties(JacksonConfiguration.PREFIX)
+public final class JacksonConfiguration {
```
graemerocher, line 41: rename to `JacksonCoreConfiguration` since we already have a type with this name in core

graemerocher (changes requested):

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonConfiguration.java
dstepanov: Renamed to `SerdeJacksonConfiguration`

graemerocher (approved):

Issue #636, opened 2023-11-02
Custom `Deserializer` for boxed type disables deserialization of primitive values


oujesky:
### Expected Behavior

```java
@Singleton
class CustomIntegerDeserializer implements Deserializer<Integer> {
    @Override
    public @Nullable Integer deserialize(@NonNull Decoder decoder, @NonNull DecoderContext context, @NonNull Argument<? super Integer> type) throws IOException {
        // ...
    }
}

@Serdeable.Deserializable
record RecordWithPrimitive(int value) {}

@Serdeable.Deserializable
record RecordWithBoxed(Integer value) {}
```

Both record should be deserializable proprely when supplying `{"value": 123}` with the value being deserialized by the custom deserializer.

### Actual Behaviour

- value with boxed type is deserialized properly
- value with primitive type throws:

```
io.micronaut.core.beans.exceptions.IntrospectionException: No bean introspection available for type [int]. Ensure the class is annotated with io.micronaut.core.annotation.Introspected
	at io.micronaut.core.beans.BeanIntrospector.lambda$getIntrospection$3(BeanIntrospector.java:116)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.micronaut.core.beans.BeanIntrospector.getIntrospection(BeanIntrospector.java:116)
	at io.micronaut.serde.support.DefaultSerdeIntrospections.lambda$getDeserializableIntrospection$1(DefaultSerdeIntrospections.java:149)
	at java.base/java.util.Optional.orElseGet(Optional.java:364)
	at io.micronaut.serde.support.DefaultSerdeIntrospections.getDeserializableIntrospection(DefaultSerdeIntrospections.java:140)
	at io.micronaut.serde.support.deserializers.ObjectDeserializer.createDeserBean(ObjectDeserializer.java:95)
	at io.micronaut.serde.support.deserializers.ObjectDeserializer.lambda$getDeserializableBean$1(ObjectDeserializer.java:87)
	at io.micronaut.core.util.SupplierUtil$2.get(SupplierUtil.java:79)
	at io.micronaut.serde.support.deserializers.ObjectDeserializer.getDeserializableBean(ObjectDeserializer.java:88)
	at io.micronaut.serde.support.deserializers.ObjectDeserializer.createSpecific(ObjectDeserializer.java:73)
	at io.micronaut.serde.support.deserializers.DeserBean.findDeserializer(DeserBean.java:567)
[...]

PR #637, opened 2023-11-02
Find custom deserializer for a primitive type

oujesky:
Solves #636

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomPrimitiveSerdeSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomPrimitiveSerdeSpec.groovy
@@ -1,0 +1,125 @@
+package io.micronaut.serde.jackson
+
+import io.micronaut.context.annotation.Primary
+import io.micronaut.context.annotation.Secondary
+import io.micronaut.core.annotation.NonNull
+import io.micronaut.core.type.Argument
+import io.micronaut.serde.*
+import io.micronaut.serde.annotation.Serdeable
+import io.micronaut.test.annotation.MockBean
+import io.micronaut.test.extensions.spock.annotation.MicronautTest
+import jakarta.inject.Inject
+import spock.lang.Specification
+
+@MicronautTest
+class CustomPrimitiveSerdeSpec extends Specification {
+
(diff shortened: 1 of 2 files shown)


serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomPrimitiveSerdeSpec.groovy
```diff
@@ -0,0 +1,125 @@
...
+        Long value
+    }
+
+    @Serdeable
+    static class TestWithPrimitiveLong {
+        long value
+    }
+
+    @MockBean
+    Serde<Integer> integerSerde() {
```
dstepanov, line 56: Why `@MockBean`? Just put `@Singleton` or `@Prototype`

dstepanov (approved):

dstepanov (changes requested):
Please try remove the moch bean

serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomPrimitiveSerdeSpec.groovy
graemerocher: maybe it is to scope the bean so it is only active for the test?

oujesky: I followed the structure that's present in [serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomStringSerdeSpec.groovy](https://github.com/micronaut-projects/micronaut-serialization/blob/master/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/CustomStringSerdeSpec.groovy), my assumption was that this scopes the bean just to the single test class, otherwise some bean conditional logic needs to be introduced

dstepanov: I see, we should have something like @TestBean for that

dstepanov: Thanks for the PR @oujesky

PR #638, opened 2023-11-02
Support `@JsonTypeInfo(visible=true)`

dstepanov:
- which allows for the deserializer to set the type discriminator as an ordinary property

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
+++ b/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
@@ -280,7 +280,7 @@ public @interface SerdeConfig {
         /**
          * Is discriminator visible.
          */
-        String DISCRIMINATOR_VISIBLE = "dvisible";
+        String DISCRIMINATOR_VISIBLE = "discriminatorVisible";
 
         /**
          * The discriminator to use.
--- a/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
+++ b/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
@@ -120,6 +120,11 @@ public @interface SerdeConfig {
     String TYPE_PROPERTY = "typeProperty";
 
     /**
+     * If the type property should be visible.
+     */
+    String TYPE_PROPERTY_VISIBLE = "typePropertyVisible";
+
+    /**
      * A property that should be used to wrap this value when serializing.
      */
     String WRAPPER_PROPERTY = "wrapperProperty";
@@ -266,10 +271,17 @@ public @interface SerdeConfig {
      */
     @Internal
     @interface SerSubtyped {
+
         /**
          * @return the subtypes
          */
         SerSubtype[] value() default {};
+
+        /**
+         * Is discriminator visible.
+         */
+        String DISCRIMINATOR_VISIBLE = "dvisible";
+
         /**
          * The discriminator to use.
          */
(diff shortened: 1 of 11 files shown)


graemerocher (approved):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
```diff
@@ -212,6 +220,15 @@ Decoder consume() {
                 }
             }
 
+            Decoder buffer() throws IOException {
```
yawkat, line 223: how can this work, you cant use the Decoder returned by decodeBuffer multiple times

serde-support/src/main/java/io/micronaut/serde/support/util/JsonNodeDecoder.java
```diff
@@ -289,10 +288,7 @@ public String decodeKey() {
 
         @Override
         public void skipValue() {
-            if (complete) {
-                throw new IllegalStateException("Already drained");
-            }
-            complete = true;
+            // No-op: Allow to read multiple times
```
yawkat, line 291: ohhhhhh. no we should definitely not allow this. it will break with more efficient implementations.

yawkat: i will try a different solution

yawkat:
ive modified it slightly, internally it's very similar (still JsonNode) but i've moved the "duplication logic" to the demuxing decoder and only apply it to the property where it's necessary. In other cases I still use normal decodeBuffer.

The reason for this change is that I want to avoid expanding the contract for decodeBuffer. Right now, decodeBuffer does not permit "aliasing": It does not allow two decodeXYZ calls to the same value, even if it's been buffered at some point. This guarantee may be helpful in the future for efficiency improvements such as lazy decoding, jackson TokenBuffer etc.

serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
```diff
@@ -266,10 +271,17 @@
...
+
         /**
          * @return the subtypes
          */
         SerSubtype[] value() default {};
+
+        /**
+         * Is discriminator visible.
+         */
+        String DISCRIMINATOR_VISIBLE = "dvisible";
```
yawkat, line 283:
```suggestion
        String DISCRIMINATOR_VISIBLE = "discriminatorVisible";
```

short name is confusing imo

yawkat (approved):

serde-support/src/main/java/io/micronaut/serde/support/util/JsonNodeDecoder.java
dstepanov: Yeah, maybe we can allow to clone the buffer

serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
dstepanov: I was trying to follow the style somebody established :D

yawkat: oh yea, the other's aren't great either imo :sweat_smile: but too late for those

altro3:
@dstepanov when you planned release these changes? Want also add fix in micronaut-openapi for version 5.2.0

dstepanov:
Probably with 4.2 platform release next week

Issue #581, opened 2023-09-26
Micronaut Serialization deserializes model bean with inheritance and discriminator property incorrectly, causing HTTP POST to REST endpoint to fail


runenielsen:
### Expected Behavior

Micronaut Serialization should set the discriminator property in an OpenAPI model bean with inheritance when deserializing.

Note that the model in the example is generated by Micronaut OpenAPI and the annotation `@JsonIgnoreProperties` is used for the discriminator with `allowSetters` set to `true`. This works fine with Jackson.

### Actual Behaviour

Micronaut Serialization does not set the discriminator property in an OpenAPI model bean with inheritance when deserializing.

This causes validation of the bean to fail, since this property is required in the specification and therefore annotated with `@NotNull`.

This is a critical bug for us, since it prevents valid models generated with Micronaut's OpenAPI generator from being posted to REST endpoints, if jakarta validation is used.

I think it is related to https://github.com/micronaut-projects/micronaut-serialization/issues/550

### Steps To Reproduce

1. Clone the repository https://github.com/runenielsen/mn-discriminator-null-serde-bug, that uses Micronaut Serialization.
2. Run the testPostModelWithDiscriminator in BookInfoControllerTest to post the model with a discriminator to a service endpoint with validation.
3. See that the test fails due to a validation failure "bookInfo.type: must not be null".
4. Now clone the repository https://github.com/runenielsen/mn-discriminator-not-null-with-jackson, that contains the same code, but uses Jackson Databind instead of Micronaut Serialization.
5. See that the same test succeeds.

### Environment Information

- MacOS
- JDK 17

### Example Application

https://github.com/runenielsen/mn-discriminator-null-serde-bug

### Version

4.1.2

graemerocher:
probably a duplicate of #550 currently `allowSetters` is unsupported and should fail with a compilation error

runenielsen:
@graemerocher : Thanks for the reply.

1. That `allowSetters` is not supported for `@JsonIgnoreProperties` seems to be missing from the [documentation](https://micronaut-projects.github.io/micronaut-serialization/latest/guide/index.html#jacksonAnnotations). There's no notes about it in the table with Jackson annotations.
2. I think this is a fairly critical missing feature, since it makes Micronaut Serialization unusable with OpenAPI specs using inheritance. It will at least prevent us from using it. Is there any plan for when you will support `allowSetters`?
[...]

Issue #550, opened 2023-08-14
Jackson's @JsonIgnoreProperties annotation not inherited


runenielsen:
### Expected Behavior

When serializing a model bean with inheritance and a discriminator property marked to be ignored with `@JsonIgnoreProperties`, I expect the discriminator property to only exist once in the JSON output by the serialization.

It should work so that any value set in the discriminator property in the bean is ignored, and the discriminator property is instead set in the serialized output based on the class of the bean. This is how it works in Jackson Databind.

### Actual Behaviour

The discriminator property exists twice in the JSON output by the serialization.

One value is generated from the value is set in discriminator property in the bean and the other value is generated from the bean's class.

### Steps To Reproduce

1. Clone the repository https://github.com/runenielsen/jsonignorepropertiesnotworking
2. Run the test JsonIgnorePropertiesWithDiscriminatorTest and see that Micronaut Serialization behaves differently than Jackson Databind.
3. I have included the test JsonIgnorePropertiesSimpleTest with a simple test of `@JsonIgnoreProperties` to verify, that it works correctly for a simple model without inheritance.
4. You can also see the incorrect output, if you run the application and open the URL http://localhost:8080/test in a browser.
5. This seems to work correctly when using `@JsonIgnore` instead of `@JsonIgnoreProperties` by the way, but the model classes in this reproduction project are a stripped down version of models generated by micronaut-openapi, so this is not something easily changeable in the class. And Micronaut Serialization should work correctly with both these annotations.

### Environment Information

MacOS
JDK 17

### Example Application

https://github.com/runenielsen/jsonignorepropertiesnotworking

### Version

4.0.3

yawkat:
this is actually unrelated to subtyping, the JsonIgnoreProperties is just not inherited from the superclass. Adding the annotation to the subclass as well works around this issue.

runenielsen:
Thanks, @yawkat 

The classes with the problem in our case is generated by Micronaut's OpenAPI generator, so it's unfortunately not easy for us to use your workaround without having to do a hack in Gradle to replace parts of them after generation.

graemerocher:
probably the fix has to go into the openapi generator to be honest

runenielsen:
@graemerocher : I know Jackson is not a standard, but I think it would be beneficial, if Micronaut Serialization works as closely as possible to Jackson, just without reflection.

[...]

PR #640, opened 2023-11-03
Support `@JsonIgnoreProperties(allowGetters,allowSetters)`, align with Jackson behaviour

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/581
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/550

### DIFF
--- a/serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java
+++ b/serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java
@@ -65,7 +65,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -687,6 +686,16 @@ public class SerdeAnnotationVisitor implements TypeElementVisitor<SerdeConfig, S
             return;
         }
 
+        if (!subtype.hasAnnotation(SerdeConfig.SerIgnored.class)) {
+            // Replicate the Jackson behaviour of using the ignore annotation from supertype but allow to override it
+            AnnotationValue<SerdeConfig.SerIgnored> serIgnored = supertype.getAnnotation(SerdeConfig.SerIgnored.class);
+            if (serIgnored != null) {
+                subtype.annotate(serIgnored);
+            }
+            // Re-evaluate ignore properties
+            visitProperties(subtype, context);
+        }
+
         final SerdeConfig.SerSubtyped.DiscriminatorValueKind discriminatorValueKind =
             getDiscriminatorValueKind(supertype);
         final SerdeConfig.SerSubtyped.DiscriminatorType discriminatorType =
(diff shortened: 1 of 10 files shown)


serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java
```diff
@@ -687,6 +686,16 @@ private void visitSubtype(ClassElement supertype, ClassElement subtype, VisitorC
             return;
         }
 
+        if (!subtype.hasAnnotation(SerdeConfig.SerIgnored.class)) {
+            // Replicate the Jackson behaviour of using the ignore annotation from supertype but allow to override it
+            AnnotationValue<SerdeConfig.SerIgnored> serIgnored = supertype.getAnnotation(SerdeConfig.SerIgnored.class);
```
yawkat, line 691: what if it's on the supersupertype

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
```diff
@@ -175,12 +175,12 @@ Entry getEntry(int i) throws IOException {
             if (buffer.size() > i) {
                 return buffer.get(i);
             } else {
-                if (buffer.size() != i) {
```
yawkat, line 178: why is this change necessary? is something calling decodeKey again after it already returned null?

serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java
dstepanov: The annotation is not inherited

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DemuxingObjectDecoder.java
dstepanov: Reverted

serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java
yawkat: in a hierarchy of A extends B and B extends C, and C has @SerIgnored, I think this will mean that A will only get SerIgnored if B is visited before A?

dstepanov: No, `@SerIgnored` is not inherited, so adding it to `B` shouldn't add it to `A`.

yawkat: but youre copying it here? i dont understand what youre saying. this code copies the annotation from the supertype, but not if it's further up, even though i assume in jackson it would still be considered for such a hierarchy

dstepanov: Well if the annotation is not inherited, putting it on B shouldn’t be present on A etc.

yawkat: but with this code, putting it on B does copy it to A.

dstepanov: The subtypes in this code are the subtypes of the base class defined by JsonSubTypes. The logic is called one. It’s not called on every type.

dstepanov: See https://github.com/micronaut-projects/micronaut-serialization/pull/640/files#diff-37b66da15700d594aac6c8e6eab4763e00f8dc3b0d769186e203b400a7a557b7R800

The ignore is not inherited. With subtypes it acts the same, if you put ignore again and want to ignore something from the supertype you need to repeat it.

yawkat: ive pushed a pendingfeature with the setup that i mean. the test passes with jackson-databind.

dstepanov: Ahh, looks like Jackson respects the last definition. I can take a look in the next PR

yawkat (approved):

Issue #642, opened 2023-11-04
`@Serializable(as = ...)` and `@Deserializable(as = ...)` doesn't check type compatibility


oujesky:
### Expected Behavior

When a type is annotated as de-/serializable with `as = ...`, there should be a check that the types must be compatible 

For example having:
```
@Serdeable.Deserializable(as = LinkedList.class)
@Serdeable.Serializable(as = LocalDate.class)
public interface SomeInterface {}
```

with `as = ...` set as completely unrelated types should throw an exception.

### Actual Behaviour

It seems that this kind of check is present in the `micronaut-serialization` annotation processor and should fail in compile time, but it looks broken.

The example code compiles successfully and then fails with an exception during both serialization (`io.micronaut.serde.exceptions.SerdeException: No serializable introspection present for type LocalDate ...` and deserialization (`io.micronaut.core.beans.exceptions.IntrospectionException: No bean introspection available for type [class java.util.LinkedList]...`).

The code that should take care of these checks is in https://github.com/oujesky/micronaut-serialization/blob/master/serde-processor/src/main/java/io/micronaut/serde/processor/SerdeAnnotationVisitor.java#L794-L812

I was tinkering around and tried debugging the `SerdeAnnotationVisitor#visitClassInternal(...)` during compilation and it found the following:
- `element.getDeclaredMetadata().findAnnotation(SerdeConfig.class)` returns `Optional.empty`
- `element.getAnnotationMetadata().findAnnotation(SerdeConfig.class)` returns `Optional[@io.micronaut.serde.config.annotation.SerdeConfig(deserAs=java.util.LinkedList, serAs=java.time.LocalDate)]`

Since declared annotation metadata are queried during the check in `SerdeAnnotationVisitor#visitClassInternal(...)`, it shows up as if the annotation wasn't present at all and the checks are skipped and silently ignored. Not sure if this is problem of the `micronaut-serialization` annotation processing or something more low-level in core.

On the other hand, during runtime, both variants return same (assumedly correct) result:
- `BeanIntrospector.SHARED.getIntrospection(SomeInterface.class).findAnnotation(SerdeConfig.class)` returns `Optional[@io.micronaut.serde.config.annotation.SerdeConfig(deserAs=java.util.LinkedList, serAs=java.time.LocalDate)]`
[...]

PR #651, opened 2023-11-07
Correct validation for `@Serializable(as = ...)` and `@Deserializable(as = ...)`

dstepanov:
+ Refactored `SerdeAnnotationVisitor` to throw an exception instead of registering a failure 

Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/642

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonGetterSetterSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonGetterSetterSpec.groovy
@@ -451,6 +451,6 @@ record Test(
 
         then:
         def e = thrown(RuntimeException)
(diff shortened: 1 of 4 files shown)


serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonGetterSetterSpec.groovy
```diff
@@ -451,6 +451,6 @@ record Test(
 
         then:
         def e = thrown(RuntimeException)
-        e.message.contains("A method annotated with AnyGetter must return a Map")
+        e.message.contains("A field annotated with AnyGetter must be a Map")
```
dstepanov, line 454: Refactoring corrected the use-case

graemerocher (approved):

PR #653, opened 2023-11-08
Fix requirement to access `ObjectSerializer` and incorrect caching argument `DeserBean`/`SerBean`

dstepanov:
Now, the leveled prefix/suffix are passed by extending the argument metadata. 

While fixing it, I discovered our key for maps of `DeserBean`/`SerBean` doesn't include annotations, and that can lead to one argument configuration being used for another one.

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/util/SerdeArgumentConf.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/util/SerdeArgumentConf.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


serde-support/src/main/java/io/micronaut/serde/support/util/SerdeArgumentConf.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.function.Predicate;
+
+/**
+ * Extra configuration placed at the argument.
+ *
+ * @author Denis Stepanov
+ * @since 2.3.2
+ */
+@Internal
+public class SerdeArgumentConf {
```
yawkat, line 41: can this be a record? might have to replace arrays with lists

yawkat (approved):

radovanradic (approved):

Issue #649, opened 2023-11-07
Serde 2.3.0 throw exception with deserialization objects with discriminator


altro3:
Exception:
```
Error decoding JSON stream for type [animal]: Unknown property [class] encountered during deserialization of type: Animal
io.micronaut.http.codec.CodecException: Error decoding JSON stream for type [animal]: Unknown property [class] encountered during deserialization of type: Animal
	at app//io.micronaut.json.body.JsonMessageHandler.decorateRead(JsonMessageHandler.java:92)
	at app//io.micronaut.json.body.JsonMessageHandler.read(JsonMessageHandler.java:106)
	at app//io.micronaut.http.netty.body.NettyJsonHandler.read(NettyJsonHandler.java:115)
	at app//io.micronaut.http.client.netty.FullNettyClientHttpResponse.convertByteBuf(FullNettyClientHttpResponse.java:225)
	at app//io.micronaut.http.client.netty.FullNettyClientHttpResponse.lambda$getBody$0(FullNettyClientHttpResponse.java:174)
	at java.base@17.0.9/java.util.HashMap.computeIfAbsent(HashMap.java:1220)
	at app//io.micronaut.http.client.netty.FullNettyClientHttpResponse.getBody(FullNettyClientHttpResponse.java:169)
	at app//io.micronaut.http.client.netty.FullNettyClientHttpResponse.<init>(FullNettyClientHttpResponse.java:108)
	at app//io.micronaut.http.client.netty.DefaultHttpClient$FullHttpResponseHandler.forwardResponseToPromise(DefaultHttpClient.java:2163)
	at app//io.micronaut.http.client.netty.DefaultHttpClient$FullHttpResponseHandler.channelReadInstrumented(DefaultHttpClient.java:2140)
	at app//io.micronaut.http.client.netty.DefaultHttpClient$FullHttpResponseHandler.channelReadInstrumented(DefaultHttpClient.java:2108)
	at app//io.micronaut.http.client.netty.SimpleChannelInboundHandlerInstrumented.channelRead0(SimpleChannelInboundHandlerInstrumented.java:46)
	at app//io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
	at app//io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
[...]

PR #656, opened 2023-11-09
Add more tests for `@JsonTypeInfo(visible = true)`

dstepanov:
Trying to reproduce https://github.com/micronaut-projects/micronaut-serialization/issues/649

@altro3 I cannot reproduce your use-case. Do you know what is different?

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonTypeInfoSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonTypeInfoSpec.groovy
@@ -1,7 +1,6 @@
 package io.micronaut.serde.jackson.annotation
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo
-import io.micronaut.context.ApplicationContextBuilder
 import io.micronaut.core.type.Argument
 import io.micronaut.serde.jackson.JsonCompileSpec
 import spock.lang.Unroll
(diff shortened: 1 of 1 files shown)


altro3:
@dstepanov Change your code in test to this and you will see Exception:
```java
package test;

import com.fasterxml.jackson.annotation.*;
import io.micronaut.core.annotation.Introspected;
import io.micronaut.serde.annotation.Serdeable;
import java.util.Optional;

@Serdeable
@JsonIgnoreProperties(
    value = "type",
    allowSetters = true
)
@JsonTypeInfo(
   use = JsonTypeInfo.Id.NAME,
   property = "type",
   visible = true)
@JsonSubTypes({
    @JsonSubTypes.Type(value = Dog.class, name = "dog"),
    @JsonSubTypes.Type(value = Cat.class, name = "cat")
})
class Animal {

    @JsonProperty("type")
    private String objType;

    public abstract String getName();

    public void setObjType(String objType) {
        this.objType = objType;
    }

    public String getObjType() {
        return objType;
    }

    @JsonIgnore
    public Optional<String> getObjTypeOptional() {
        return Optional.ofNullable(objType);
    }

}

class Dog extends Animal {

    public double barkVolume;
    private String name;

    @Override
    public String getName() {
        return name;
    }
    public void setName(String name) {
        this.name = name;
    }
}

class Cat extends Animal {

    public boolean likesCream;
    public int lives;
    private String name;

    public String getName() {
        return name;
    }
    public void setName(String name) {
        this.name = name;
    }
}
```

altro3:
@dstepanov I think now the problem here:
```java
@JsonSubTypes({
    @JsonSubTypes.Type(value = Dog.class, name = "dog"),
    @JsonSubTypes.Type(value = Cat.class, name = "cat")
})
```

This is second way to declare subtypes. You used `@JsonTypeName` before

dstepanov:
It works for me, your code has compilation problems but if I fixed them it's passing

PR #657, opened 2023-11-09
Problem with `Must access entries in sequence`

dstepanov:
@yawkat Can you please take a look at this case? I think we should allow multiple called to `decodeKey` after it returns null.


yawkat:
@dstepanov the reason i do not want to allow that is because Decoder does not require more than one decodeKey==null to succeed. In my mind, decodeKey advances the parser, and advancing the parser twice is not valid. I think it's a bug in the caller, like you can see in my fix it's possible to write deserializers in such a way that decodeKey is not called multiple times. 

However all the other implementations of decodeKey handle multiple calls just fine. so i could be convinced to relax this requirement. but if we do it should be documented in Decoder.

yawkat:
fwiw if this becomes a common issue, i think we should relax the requirement.

PR #658, opened 2023-11-09
Added test for #649

altro3:


altro3:
@dstepanov I've added full classes structure for test

altro3:
@dstepanov hmm... very strange, I see that test works fine.... Maybe this mapper is not the same as used in micronaut controllers?

dstepanov:
Maybe the annotation processor is different

altro3:
@dstepanov look to this. I debugged micronaut-openapi test. And I see, that deserialization works fine. But! Problem in this line
```java
final DeserBean.DerProperty<Object, Object> consumedProperty = propertiesConsumer.consume(prop);
```

consumedProperty is null for discrimantor. Look to the screenshot

изображение

altro3:
@dstepanov As I see, The problem is that the class field is trying to be processed twice even though it has already been processed

graemerocher:
we need a reproducer

altro3:
@dstepanov Also I see that the response body deserialized twice. First time here `ImmediateByteBody.processSingle(...)`
изображение

Second time here: `FullNettyClientHttpResponse.convertByteBuf`
изображение

altro3:
@graemerocher To reproduce the problem, you need to lift the entire bunch to work with micronaut. Those. We need a controller, a client and a serde. As far as I understand, the problem is not serialization, but somewhere in the connection between the client and serde

graemerocher:
@yawkat can you look into the double deserialisation issue?

altro3:
@graemerocher I'll try to crete a sample reproducer tomorrow or on the weekend

yawkat:
@graemerocher one of the deserializations happens on the server, one on the client. i dont think it's an issue.

altro3:
@yawkat yes, sorry. I realized this when I was already in bed :-).

@graemerocher @dstepanov Reproducer is ready. Here: https://github.com/altro3/micronaut3-bug

To reproduce problem just satrt Application and call endpoint https://localhost:8080/start and you will see this exception.

I was able to identify the problem: the problem is not with the client, but with incorrect serialization. The class field is duplicated in the final json. To see this, change `line 47` in the controller with this:
```java
var resposne = httpClient.retrieve(HttpRequest.<Animal>POST("/test", bird), Animal.class);
```
To this line:
```java
var resposne = httpClient.retrieve(HttpRequest.<Animal>POST("/test", bird), String.class);
```
And you will see what the controller returns to us:
```java
[MyEntityController.java:56] - Response: {"class":"ave","numWings":2,"beakLength":1.2,"featherDescription":"Large blue and white feathers","class":"ave","color":"blue"}
```
As you see field `"color":"blue"` contatins in response twice.

dstepanov:
I suspect the plugin will override the annotation processor. Might need to force the dependency

altro3:
@dstepanov No, the problem not in serde version

изображение

altro3:
And what I see after build:
изображение
serde libraries with correct version

PR #708, opened 2023-12-11
Fix JsonProperty on mixin methods

yawkat:
Do you remember why you added this check @sdelamo?


dstepanov (approved):

sdelamo (approved):
@graemerocher added that in this commit https://github.com/micronaut-projects/micronaut-serialization/pull/447/commits/70cd005f6c5795fe95a280f06e9d9bcb293066fb to fix the upgrade to Micronaut Framework 4.

PR #711, opened 2023-12-12
Apply micronaut.serde.time-zone for ISO-8601 DateTimeFromatter in DefaultFormattedTemporalSerde

VladM-N1:
Hi everyone, this is my PR to any open-source project ever.
I believe this change more like QoL.
Problem:
`micronaut.serde.time-zone` configuration only applies to custom configured `DateTimeFormatter` which can be created only with: `micronaut.serde.date-format`.


yawkat:
i dont think we should make this change. the time zone should not be adjusted by the serializer. i dont know why we support this for serialization in the first place, it should only apply for deserialization as a default tz.

yawkat:
I think it makes some sense for deserialization. my guess is that the serialization behavior was an accident.

PR #712, opened 2023-12-12
Field marked with access READ_ONLY should not be serialized

radovanradic:


serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/SerdeJsonPropertySpec.groovy
```diff
@@ -254,4 +254,41 @@ record Test(
...
 
+    void "object with readonly field"() {
+        given:
+        def context = buildContext('example.Test', '''
+package example;
+
+import com.fasterxml.jackson.annotation.*;
+@io.micronaut.serde.annotation.Serdeable
+class Test {
+    private String foo;
```
radovanradic, line 265: @dstepanov Looks like something got changes in 2.5.0 and this annotation is being ignored when filtering write properties in SerBean.

dstepanov: Actually, our logic was incorrect, READ_ONLY = serialization only. You can try to add this test into the superclass and test it with Databind

radovanradic: How we can annotate field to be just read only? It is not being serialized but can be read from the repository (like etag or version from the database)?

dstepanov: Only deserialized is WRITE_ONLY

radovanradic: It's awkward because jsonMapper.writeValue won't write field and it is WRITE_ONLY. And jsonMapper.readValue will read it and it is WRITE_ONLY. But ok, I will try that.

radovanradic: Yes, passing now

radovanradic:
Not needed, need to configure field WRITE_ONLY instead of READ_ONLY

serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/SerdeJsonPropertySpec.groovy
dstepanov: Yeah but that's coming from Jackson

PR #734, opened 2024-01-23
Fix customized map serializer

radovanradic:
Found this bug when tried to serialize object with the map


serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedMapSerializer.java
```diff
@@ -82,7 +82,7 @@ public boolean isEmpty(EncoderContext context, Map<K, V> value) {
                 public void serialize(Encoder encoder, EncoderContext context, Argument<? extends Map<K, V>> type, Map<K, V> value) throws IOException {
                     // slow path, lookup each value serializer
                     final Encoder childEncoder = encoder.encodeObject(type);
```
radovanradic, line 84: Probably used encoded instead of childEncoder. Without this change, serializing would throw error
```
java.lang.IllegalStateException: Already have a key
	at app//io.micronaut.serde.support.util.JsonNodeEncoder$Obj.encodeKey(JsonNodeEncoder.java:176)
```

graemerocher (approved):

Issue #690, opened 2023-11-21
Deserializer for a collection with specific type breaks default generic collection deserializer


oujesky:
### Expected Behavior

- Imagine a JSON payload that contains a string value with several actual delimited values (e.g. `{"something": "a|b|c"}`
- To deserialize this to a `List<Something>` a custom deserializer `Deserializer<List<Something>>` is created.  
- At the same time, there might be a actual JSON array that is deserialized to a `List<X>` using one of the built-in default collection deserializer (i.e. `ArrayListDeserializer`)

### Actual Behaviour

- When a `Deserializer<List<Something>>` is defined, it is picked up by `DefaultSerdeRegistry#findDeserializer(...)` for all other `List<X>` deserializations instead of the default deserializer. 

### Steps To Reproduce

1. Create a deserializable POJO/record 
```java
@Deserializable
record SomeModel(List<Something> specificList, List<String> genericList) {}
```
2. Create a custom deserializer for `List<Something>`
```java
@Singleton
class ListSomethingDeserializer implements Deserializer<List<Something>> {
    @Override
    public @Nullable List<Something> deserialize(@NonNull Decoder decoder, @NonNull DecoderContext context, @NonNull Argument<? super List<Something>> type) throws IOException {
        var stringValue = decoder.decodeString();
        return Arrays.stream(stringValue.split("\\|"))
            .map(Something::new)
            .toList();
    }    
}
```
3. Deserialize JSON payload
```json
{
    "specificList": "a|b|c",
    "genericList": ["a", "b", "c"]
}
```
using 
```java
var result = objectMapper.readValue(json, SomeModel.class);
```
5. The `List<Something>` is deserialized using the custom deserializer, but deserializing the `List<String>` is also trying to use the custom deserializer and fails with an exception (that would vary with the nature of the data and its custom deserializer), in this case `io.micronaut.serde.exceptions.SerdeException: Error decoding property [List genericList] of type [class com.example.SomeModel]: Expected one string, but got array of multiple values` with a stack trace leading to the custom deserializer.

This is captured in linked reproducer repo: https://github.com/oujesky/micronaut-serde-wrapper-type/blob/main/src/test/java/com/example/MicronautSerdeWrapperTypeTest.java

### Environment Information

- JDK 17
- macOS

### Example Application

https://github.com/oujesky/micronaut-serde-wrapper-type

### Version

4.2.0

PR #742, opened 2024-01-30
Implement correct generic arguments matcher

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/690


graemerocher:
probably should run this through the Micronaut Data test suite to see if any coverage is missing here

Issue #754, opened 2024-02-03
`@JsonValue` doesn't work on property level


altro3:
See test here : https://github.com/micronaut-projects/micronaut-serialization/pull/756

altro3:
@dstepanov @graemerocher Still doesn't work :-(

Related issue: https://github.com/micronaut-projects/micronaut-serialization/issues/168

altro3:
allw orks fine. Sorry, my mistake

altro3:
@dstepanov I found a strange test for this case that doesn't actually test anything.:
![изображение](https://github.com/micronaut-projects/micronaut-serialization/assets/48474279/9de959b2-8403-4fed-adeb-a585254e5b27)

I added a full-fledged test and indeed, databind works correctly, serde does not understand that the minimum value should be different

PR #756, opened 2024-02-04
Add test for `@JsonValue` annotation on property level

altro3:
See this: #754


dstepanov:
Is it passing?

dstepanov:
This is fixed already

PR #761, opened 2024-02-07
Convert internal serdes from beans to ordinary classes

dstepanov:
This should improve the startup.

I have extracted all of the serdes defined in the `DefaultSerdeRegistry` into own package.

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/DefaultSerdeRegistry.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/DefaultSerdeRegistry.java
@@ -20,6 +20,7 @@ import io.micronaut.context.BeanResolutionContext;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.context.annotation.Secondary;
 import io.micronaut.context.exceptions.ConfigurationException;
+import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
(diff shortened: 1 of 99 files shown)


serde-support/src/main/java/io/micronaut/serde/support/serdes/InstantSerde.java
```diff
@@ -27,8 +28,7 @@
  *
  * @since 1.0.0
  */
-@Singleton
-public final class InstantSerde extends NumericSupportTemporalSerde<Instant> implements TemporalSerde<Instant> {
+public final class InstantSerde extends NumericSupportTemporalSerde<Instant> implements TemporalSerde<Instant>, SerdeRegistrar<Instant> {
```
graemerocher, line 31: Will this break existing use cases? We do this for example:

```
@Singleton
@Primary
final class KievInstantSerde implements Serde<Instant> {
    private final InstantSerde instantSerde;

    public KievInstantSerde(InstantSerde instantSerde) {
        this.instantSerde = instantSerde;
    }

    @Override
    public @Nullable Instant deserialize(@NonNull Decoder decoder, @NonNull DecoderContext context, @NonNull Argument<? super Instant> type) throws IOException {
        if (decoder instanceof KievDecoder kievDecoder) {
            Timestamp timestamp = kievDecoder.decodeTimestamp();
            return timestamp.toInstant();
        } else {
            return instantSerde.deserialize(decoder, context, type);
        }
    }
```

How will this work with this change?

graemerocher (commented):
Are there some JMH benchmarks that show the improvement in startup time vs the old way?

serde-support/src/main/java/io/micronaut/serde/support/serdes/InstantSerde.java
dstepanov: Ok this needs to be corrected

dstepanov:
I don't see much of the improvement. But having fewer beans might help.

serde-support/src/main/java/io/micronaut/serde/support/DefaultSerdeRegistry.java
```diff
@@ -255,32 +387,69 @@ public SerdeRegistry cloneWithConfiguration(@Nullable SerdeConfiguration configu
...
+     *
+     * @param type The serde type
+     * @param <T>  The serde type
+     * @return a serde or null
+     */
+    @Nullable
+    @Internal
+    public <T> Serde<T> findInternalSerde(Argument<T> type) {
+        for (BeanDefinition<Serde> serdeBeanDefinition : internalSerdes) {
+            if (serdeBeanDefinition instanceof InternaSerdeBeanDefinition<?> internaSerdeBeanDefinition
```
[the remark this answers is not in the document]
dstepanov: 🤦‍♂️

yawkat (approved):

PR #771, opened 2024-02-19
Enable native examples and the server TCK

dstepanov:


graemerocher (approved):

sdelamo (approved):

sdelamo (changes requested):
native tests seems to fail

dstepanov:
I had to disable the static test; it's also disabled in the servlet module.

yawkat:
it seems like the build could be fixed to include the static resources

dstepanov:
How?

sdelamo (approved):

Issue #170, opened 2022-03-27
Add comment support to json


altro3:
### Feature description

This is not a common case, but sometimes when you use json for the configuration format, you need to make some notes inside the json . I know the json standard doesn't support comments, but for example jackson has a feature to deserialize json with comments.

There is currently no way to configure objectMapper in serde to support comments. I also think that it is necessary to add the ability to create a custom objectMapper with non-typical settings, for example, support for comments :-).

altro3:
@graemerocher @dstepanov @sdelamo Any news about this ? 

You need to understand why this feature is extremely necessary:

Without it, it is impossible to use the json format for configuration normally. Of course, comments are required in large configurations. otherwise, it is impossible to work normally with such a configuration. If anything, jackson supports comments in json

dstepanov:
You can configure Jackson read/write features see `SerdeJacksonConfiguration`

sdelamo:
As you said, JSON does not support comments in the specification. I would not complicate the code to support a not standard feature. 

Moreover, we are not recommending people to use JSON for configuration but formats such as properties, yaml, toml. Formats which support comments.

altro3:
The decision on the choice of format does not always depend on you. Yes, I understand that the specification does not provide for comments in json. But all popular solutions for working with json support the inclusion of comments in the parser, I mean jackson and gson.

As I understand from the project, this is an attempt to speed up the work of these libraries at the expense of the mikronaut code generator. If so, then it is quite logical to support all the functionality of these libraries so that it is really possible to connect micronaut-serde.

So, if I were you, I would not focus on specifications or other bureaucratic inventions, but rather look at what is actually used by real programmers

My opinion is that you should create not on the basis of rules invented by someone, but on the basis of logic, common sense and efficiency. And if the rules get in your way, then your task is to expand the rules.

Programs and computers work based on rules. So who are we - programs or real people?

graemerocher:
@altro3 did you try @dstepanov's recommendation of setting the Jackson feature to allow comments ?https://fasterxml.github.io/jackson-core/javadoc/2.8/com/fasterxml/jackson/core/JsonParser.Feature.html#ALLOW_COMMENTS

graemerocher:
See https://micronaut-projects.github.io/micronaut-serialization/latest/guide/configurationreference.html#io.micronaut.serde.jackson.SerdeJacksonConfiguration

graemerocher:
[...]

PR #778, opened 2024-02-27
Allow to clone `JacksonJsonMapper` with a different `SerdeJacksonConfiguration`

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/170


graemerocher (approved):

yawkat:
do we need the interface

dstepanov:
I would prefer to have an interface instead of exposing `JacksonJsonMapper` to be public.

yawkat:
ok

yawkat (approved):

PR #794, opened 2024-03-18
Bug: empty string serde and Kotlin in 2.8.2

timyates:
I believe there is an issue in 2.8.2 when serializing an empty string to a non-null field in Kotlin, which is reproduced by this PR.

Instead of the expected validation error `fruitCommand.name: must not be empty`, with Kotlin we get:

```
Parameter specified as non-null is null: method example.FruitCommand.<init>, parameter name
```

This works in 2.8.1, and was exposed by the eclipsestore guide https://github.com/micronaut-projects/micronaut-guides/pull/1443\#commits-pushed-0276ab1


timyates:
Actually, I'm not sure if this reproducer is correct...

Going back to earlier commits, the Kotlin tests in this PR still seem to fail, and I'm not sure why...

However in the EclipseStore guide (check out micronaut-guides, and run `./gradlew mEPBuild`), adding

```
configurations.all {
    resolutionStrategy.eachDependency { details ->
        if (details.requested.group == "io.micronaut.serde") {
            details.useVersion("2.8.1")
        }
    }
}
```

To the build of the kotlin guide does make it pass...

dstepanov:
It looks correct to me; the default inclusion setting is NON_EMPTY. So, an empty string is omitted.
For Java we can initialize beans even with non-null but not for Kotlin.

timyates:
@dstepanov did something change to make the old way invalid?  As far as I can see, this used to work

dstepanov:
Not sure, maybe the configuration of the inclusion wasn't properly delegated?

dstepanov:
I think there was additional String Serde that didn't had implemented isEmpty method https://github.com/micronaut-projects/micronaut-serialization/commit/6489c8eec86ae499e7c8dcfdd51c7ee188390db6#diff-56e4febbd38eabfee4eb5f5829aa0e2a3b0a9af1d7be8dcb75bb6dd8673b9ae4L183

PR #871, opened 2024-06-20
`JacksonJsonMapper` should support `createSpecific`

dstepanov:


yawkat (approved):

dstepanov:
Looks like we are trying to resolve mappers for void :(
`io.micronaut.core.beans.exceptions.IntrospectionException: No bean introspection available for type [void]`
In a case of the exception I will just return the non-specific mapper

Issue #768, opened 2024-02-15
KSP: Serde can't work with discrimantor


altro3:
Get a sample application: https://github.com/altro3/micronaut3-bug/tree/kotlin (branch `kotlin`)

Send simple request to endpoint `localhost:8080/test`
```json
{
    "numWings": 2,
    "beakLength": 12.1,
    "featherDescription": "this is description",
    "class": "ave",
    "color": "red"
}
```

And you will see exception: 

![изображение](https://github.com/micronaut-projects/micronaut-serialization/assets/48474279/d8ff40ed-e14a-4be8-93d0-8cdd34cff2f0)

Sample controller:
```kotlin
    @Post("/test")
    open fun test(@Body @NotNull @Valid animal: Animal): Mono<Animal> {
        return Mono.just(animal)
    }
```

Sample model classes:
```kotlin
@Serdeable
@JsonPropertyOrder(
        Animal.JSON_PROPERTY_PROPERTY_CLASS,
        Animal.JSON_PROPERTY_COLOR
)
@JsonIgnoreProperties(
        value = ["class"], // ignore manually set class, it will be automatically generated by Jackson during serialization
        allowSetters = true // allows the class to be set during deserialization
)
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "class", visible = true)
@JsonSubTypes(
        JsonSubTypes.Type(value = Bird::class, name = "ave")
)
open class Animal(
        @Nullable
        @JsonProperty(JSON_PROPERTY_COLOR)
        @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
        open var color: ColorEnum? = null,
        @Nullable
        @JsonProperty(JSON_PROPERTY_PROPERTY_CLASS)
        @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
        open var propertyClass: String? = null,
) {

    companion object {

        const val JSON_PROPERTY_PROPERTY_CLASS = "class"
        const val JSON_PROPERTY_COLOR = "color"
    }
}
```
Bird
```kotlin
@Serdeable
@JsonPropertyOrder(
        Bird.JSON_PROPERTY_NUM_WINGS,
        Bird.JSON_PROPERTY_BEAK_LENGTH,
        Bird.JSON_PROPERTY_FEATHER_DESCRIPTION
)
data class Bird(
        @Nullable
        @JsonProperty(JSON_PROPERTY_NUM_WINGS)
[...]

PR #875, opened 2024-06-21
Add test for process KSP Json subtypes. See #768

altro3:
See https://github.com/micronaut-projects/micronaut-serialization/issues/768


altro3:
@dstepanov Are you planned to fix it? I still have some tests disabled in micronaut openapi for KSP because of this bug :-(

dstepanov:
@altro3 It works correctly, you need to have `visible = false` because the bird class doesn't have the class property

altro3:
@dstepanov Then I have a question: why does this code work correctly with KAPT and java, but not only with KSP?

altro3:
@dstepanov You are wrong. The `Bird` class inherits from the `Animal` class. The `Animal` class has a property `class`
{2A9FBF20-8F02-418D-8B2D-7758EA83907F}

altro3:
@dstepanov Tets should pass exactly with the configuration as I wrote. I think that KSP does not see the property `class`, because it is in the parent class

altro3:
I just want to say: if a property is not in the constructor, it does not mean that the parent does not have the property. Yes, you will not be able to set values ​​for these properties in the constructor, but you will be able to specify them additionally. In our case, like this:

{B3F1CD85-F665-451A-9261-602FE6906323}

dstepanov:
Needs `@field:JsonProperty(JSON_PROPERTY_PROPERTY_CLASS)`

altro3:
Well, that's a bug. Why do all other annotations work fine without the "field:" prefix, but only in this case does it not work without the prefix?

altro3:
@dstepanov As you see, in case withou prefix `@filed:`,  `@JsonProperty` annotation was processed and `DeserBean` class found right information about discriminator property:
{57C97479-AB80-49C2-A101-F22EF01FDD7D}

But why does the processing continue incorrectly?

altro3:
I think I understand what the problem is and it is clearly in the lkotlin introspection, that is, in the core module.

In this case, the annotations specified in the superclass construct - Animal are ignored, but these annotations are also valid for fields. This was, among other things, the point of the KSP, to get rid of these stupid "field:" prefixes.

altro3:
In short, it is necessary that the annotations specified in the superconstructor are added to the annotations of the fields. Otherwise, you get a hodgepodge - in one case you can omit the "field" prefix, and in others - not.

dstepanov:
Good question, I think the annotation in the default variant is only added to the constructor and we only take field, getter and setter annotations to the final annotation metadata. You can debug and see. Kapt probably is putting in everywere in the stub

altro3:
Yes, I understand what the problem is. Question for you: can this be fixed, or should we just give in and shut up? :-)

altro3:
@dstepanov I decided to test another case: if we do not redefine class attributes, but simply pass arguments of parent constructors in all child constructors. This way we are guaranteed to always use a constructor, and not setters for individual attributes.

Look at the test structure:
```kotlin
@Serdeable
@JsonIgnoreProperties(
    value = ["myType"], // ignore manually set type, it will be automatically generated by Jackson during serialization
    allowSetters = true, // allows the type to be set during deserialization
)
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "myType", visible = true)
@JsonSubTypes(
    JsonSubTypes.Type(value = BasicBookInfo::class, name = "BASIC"),
    JsonSubTypes.Type(value = DetailedBookInfo::class, name = "DETAILED"),
)
open class BookInfo(

    var name: String,

    @JsonProperty("myType")
    var type: BookInfoType? = null,
)

@Serdeable
open class BasicBookInfo(

    @NotNull
    @Size(min = 3)
    var author: String,

    name: String,
    type: BookInfoType? = null,

) : BookInfo(name, type)

@Serdeable
class DetailedBookInfo(

    @Pattern(regexp = "[0-9]{13}")
    var isbn: String,

    author: String,
    name: String,
    type: BookInfoType? = null,

) : BasicBookInfo(author, name, type)
```

As you can see, the type attribute has the `@JsonProperty` annotation, but it is only specified in the constructor argument of the main parent.

So, why is this annotation ignored when introspection occurs?

Yes, of course, I can specify the `field` prefix and everything will work, but the idea is that constructor annotations are taken into account throughout the class hierarchy.

I believe this is a bug

dstepanov:
Well, `JsonProperty` is not inherited, so it should not appear on the inherited parameter.

Looking at `AbstractKotlinPropertyElement` and `PropertyElementAnnotationMetadata` I think we need to merge the annotations added to the property. The concept that doesn't exist in Java so that's why it's happening.

altro3:
Well, look what I got:

1. I improved the generator, taking into account your comment about the field prefix

2. Also, the generator now duplicates the annotations of the constructor arguments in each child class - this solves the second problem.

After these changes, the tests in this PR began to pass correctly

Now let's get back to the test in micronaut-openapi. And the test still doesn't pass! And here's the problem: in the "class" field it writes the class name (!!!!!), although there should be a discriminator value!!!

But it looks like the problem is not in the micronaut-serde library, but in the http client!

Can you help me figure out this problem? It only occurs when using KSP and the problem is definitely not in the annotations, since they are now guaranteed to be set correctly

This is a screenshot showing the problem. Look at the value of the propertyClass field - the class name is `Bird`, but it should be `ave` !

326349724-501fe05f-4260-4dcd-af12-a5ba24fd7cd6

dstepanov:
What do the bean's definitions look like?

altro3:
Look, the problem is NOT in the HTTP client, but in the controller!!!

It looks like there is a serialization and deserialization error somewhere, and the real class name is automatically substituted into the discriminator value field!

I really don't understand where the "substitution" of the discriminator value for the class name occurs... In the test in the current PR, I specifically made a similar model, as in my test - but in this PR everything works correctly... It turns out that the problem is in micronaut http

Get sources fro mthis branch: https://github.com/altro3/micronaut-openapi/tree/reactive-tests

Run test in module `test-suite-kotlin-ksp-server-generator` run test `RequestBodyControllerTest.testSendModelWithDiscriminatorChild1`

1. look to the request model:
изображение

`propertyClass` field is null, and we have `Bird` object

2. In controller we see this:

{27C800E6-56A4-49A2-88A7-F5B17AFE67C7}

As you see in `propertyClass` field we have value `Bird`, but must be `ave`

altro3:
Here you will see all generated controller interfaces and data models

{4B807F2E-11C8-49B5-9CD1-9AAC24F87294}

altro3:
With java and KAPT this test works fine

altro3:
@dstepanov I started debugging the process of sending a request in the test. This is what I see

1. The problem really occurs when serializing the request in the client.

2. The screenshot shows that the injected value is incorrect and corresponds to the class name, and not the value from the annotations

It is also clear that in the object itself this field is empty, that is, it is injected by micronaut-serde .

I suspect that everything is connected with the name of the field - `class` and if this field is called something else, for example, `type`, then everything will work. But this is only an assumption. which I have not yet checked

So, look at the screenshot with KSP:
{3CD63A05-9000-4282-B9CB-0257E49DCEFE}

The same checkpoint with KAPT:
{16292EE4-DBB8-489C-95B5-67AE97C6B973}

Look, the `injected` value is correct - `ave` . So now I'm 100% sure that the bug is in micronaut-serialization introspection

altro3:
@dstepanov I think, found it. Looks to `$Bird$Introspection.class` (left - KAPT, right - KSP):

{A6C0C6C0-ACAD-4309-A323-4B94174F2EF2}

altro3:
@dstepanov Added correct test to reproduce this problem: `example.RequestBodyControllerTest`

Please, help to fix it!

dstepanov:
The main problem is that KSP doesn't allow fetching the annotation model from a string and reading the annotations. We need to get the repeatable container. 
There is a trick to get the class if it's on the classpath, but it looks like KSP has some classloading mishmash.

https://github.com/micronaut-projects/micronaut-core/pull/11408

I can make the build pass if I add:

```
        Thread.currentThread().setContextClassLoader(visitorContext.getClass().getClassLoader());
```

in `JsonSubTypesMapper`

altro3:
@dstepanov This fix doesn't seem to help :-(

PR #897, opened 2024-07-23
Add type thrashing test

yawkat:
Simple test that does back-and-forth serialization and looks for type thrashing.

Also contains a fix for the DefaultSerdeRegistry cache map to avoid the thrashing. Other than that this seems to be a non-issue.


dstepanov (approved):

dstepanov:
Maybe we can have some simple beans tests

graemerocher (approved):

yawkat:
STRING_ARRAY_FIELD is a bean with a single field. not comprehensive but it tests the basics.

PR #898, opened 2024-07-23
Fix JsonNumber bson serialization

rorueda:
- Always use the right number type when serializing, so formats that support the type info can deserialize to the correct type. Currently, if you serialize an `int JsonNumber`, when deserializing you get a `long JsonNumber`.
- Fix fallback deserialization of custom `Number` implementations. The bson decoder advances when `getBigDecimal` is called - so, in the second call you either get an exception or one of the fields is swallowed.
- Fix element index increment when serializing `BigInteger`. Increment is already done by `encodeBigDecimal`.

### DIFF
--- a/serde-bson/src/main/java/io/micronaut/serde/bson/BsonReaderDecoder.java
+++ b/serde-bson/src/main/java/io/micronaut/serde/bson/BsonReaderDecoder.java
@@ -280,6 +280,11 @@ public final class BsonReaderDecoder extends AbstractDecoderPerStructureStreamDe
     }
 
     @Override
+    protected BigDecimal getBigDecimalFromNumber(Number number) {
+        return ((Decimal128) number).bigDecimalValue();
+    }
+
+    @Override
     public byte @NonNull [] decodeBinary() throws IOException {
         if (currentBsonType == BsonType.BINARY) {
             return decodeCustom(parser -> ((BsonReaderDecoder) parser).bsonReader.readBinaryData().getData());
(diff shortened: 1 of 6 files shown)


serde-bson/src/main/java/io/micronaut/serde/bson/BsonReaderDecoder.java
```diff
@@ -279,6 +279,11 @@ protected Number getBestNumber() {
         };
     }
 
+    @Override
+    protected BigDecimal getBigDecimalFromNumber(Number number) {
+        return ((Decimal128) number).bigDecimalValue();
```
dstepanov: This can be implemented using instanceOf or different convertions

dstepanov (changes requested):

serde-support/src/main/java/io/micronaut/serde/support/AbstractStreamDecoder.java
```diff
@@ -402,6 +402,19 @@ protected int getInteger() throws IOException {
...
+     * Converts the number, probably retrieved by calling {@link #getBestNumber()}, to a {@link BigDecimal},
+     * when it is not one of {@link Byte}, {@link Short}, {@link Integer}, {@link Long}, {@link Float},
+     * {@link Double}, {@link BigInteger}, {@link BigDecimal}
+     *
+     * @param number The number value
+     * @return The number as a big decimal
+     * @throws UnsupportedOperationException If custom number types are not expected
+     */
+    protected BigDecimal getBigDecimalFromNumber(Number number) {
+        throw new UnsupportedOperationException();
```
yawkat: Definitely needs a proper default implementation.

dstepanov:
Thanks for the PR. I have corrected the new method.

PR #943, opened 2024-10-08
Implement JsonView support

yawkat:
Complementary to https://github.com/micronaut-projects/micronaut-core/pull/11189 .


serde-support/src/main/java/io/micronaut/serde/support/util/ViewUtil.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    @Contract(pure = true, value = "_, _, !null -> !null")
+    public static Class<?> extractView(@Nullable SerdeConfiguration configuration, @NonNull Argument<?> argument, Class<?> defaultValue) {
+        if (configuration != null && configuration.isJsonViewEnabled()) {
+            AnnotationMetadata annotationMetadata = argument.getAnnotationMetadata();
+            Optional<Class> jackson = annotationMetadata.classValue("com.fasterxml.jackson.annotation.JsonView");
+            if (jackson.isPresent()) {
+                return jackson.get();
+            }
+            Optional<Class> serde = annotationMetadata.classValue(SerdeConfig.class, SerdeConfig.VIEWS);
```
dstepanov: We don't remap to `SerdeConfig.VIEWS`?

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonJsonMapper.java
```diff
@@ -128,8 +128,9 @@ private JacksonJsonMapper(@NonNull SerdeRegistry registry,
     @NonNull
     @Override
     public JsonMapper createSpecific(@NonNull Argument<?> type) {
+        JacksonJsonMapper mapper;
```
dstepanov, line 131: It looks to me that having the view in the mapper is useless. 
I think it should be extracted in `ObjectDeserializer/ObjectSerializer` and create a new serializer /deserializer wrapper that will create a new context with something like `context = context.withViews(...)`

but that will not add a view for custom serdes so I don't know.

serde-api/src/main/java/io/micronaut/serde/config/DefaultSerdeConfiguration.java
```diff
@@ -59,7 +61,8 @@ final class DefaultSerdeConfiguration implements SerdeConfiguration {
                               @Bindable(defaultValue = "io.micronaut") List<String> includedIntrospectionPackages,
                               @Bindable(defaultValue = LimitingStream.DEFAULT_MAXIMUM_DEPTH + "") int maximumNestingDepth,
                               @Bindable(defaultValue = DEFAULT_INET_ADDRESS_AS_NUMERIC + "") boolean inetAddressAsNumeric,
-                              @Nullable String propertyNamingStrategy) {
+                              @Nullable String propertyNamingStrategy,
+                              @Property(name = "jackson.json-view.enabled", defaultValue = "false") boolean jsonViewEnabled) {
```
dstepanov: Please add a `micronaut.serde.json-view.enabled` alternative and a note in the documentation.

serde-support/src/main/java/io/micronaut/serde/support/util/ViewUtil.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Optional;
+
+/**
+ * Utilities for implementing JsonView support in mappers.
+ *
+ * @author Jonas Konrad
+ * @since 2.12
+ */
+@Internal
+public final class ViewUtil {
```
graemerocher: rename to `JsonViewUtil`

dstepanov (changes requested):

serde-api/src/main/java/io/micronaut/serde/Deserializer.java
```diff
@@ -128,7 +128,7 @@ default ConversionService getConversionService() {
          * @return {@code true} iff any of the given views is enabled.
          */
         default boolean hasView(Class<?>... views) {
-            return false;
+            return true;
```
graemerocher, line 131: why was this default changed to `true`?

serde-api/src/main/java/io/micronaut/serde/Serializer.java
```diff
@@ -108,7 +108,7 @@ default ConversionService getConversionService() {
          * @return {@code true} iff any of the given views is enabled.
          */
         default boolean hasView(Class<?>... views) {
-            return false;
+            return true;
```
graemerocher, line 111: why was this default changed to `true`?

graemerocher (changes requested):

serde-api/src/main/java/io/micronaut/serde/Deserializer.java
yawkat: Jackson databind behavior is that if no view is specified, all fields are included.

serde-support/src/main/java/io/micronaut/serde/support/util/ViewUtil.java
yawkat: We do, but I want to support both because we don't usually require the serde processor on controllers, so JsonView may not always be remapped

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonJsonMapper.java
yawkat: i think it is sensible that the view is in the encoder context. i dont think a serializer wrapper is the right place to modify the context depending on view, the fact that we change the context view based on the argument is very specific to the JsonMappers so that is where it should stay. If we included this extractView logic inside a serializer, we risk using the JsonView of a property to modify the encoder context, which is not what a JsonView on a property is supposed to do (it is supposed to modify the visibility of the property, not change the view for child properties). And of course there is the technical issue that we'd need to change this for every createSpecific implementation unless we start wrapping literally every serde.

graemerocher (approved):

dstepanov (approved):

PR #955, opened 2024-10-24
Unwrapped/nested properties don't work with Builder

radovanradic:


serde-jackson-tck/src/main/groovy/io/micronaut/serde/jackson/JsonUnwrappedSpec.groovy
```diff
@@ -957,4 +957,56 @@ class SubClass extends SuperClass {
...
+""")
+
+        when:
+        def nestedJsonStr = '{"value":"test1","addr_street":"Blvd 11","addr_city":"NY"}'
+        def deserNestedEntity = jsonMapper.readValue(nestedJsonStr, Argument.of(ctx.classLoader.loadClass('unwrapped.test.TestNestedEntity')))
+
+        then:
+        deserNestedEntity
+        deserNestedEntity.value == "test1"
+        deserNestedEntity.address.city == "NY"
```
radovanradic, line 1005: Looks like nested fields are not deserialized when using builder in `@Introspected`

dstepanov (approved):

PR #981, opened 2024-11-26
Add serde for jackson-databind JsonNode

yawkat:
I added it to the support module so that it still works with a non-jackson ObjectMapper.


dstepanov (approved):
Do we need it?

yawkat:
apparently some people do

PR #986, opened 2024-12-03
Build time initialize DiscriminatorValueKind inner enum

andriy-dmytruk:
Since `DiscriminatorValueKind` enum is used as annotation value, the class needs to be build-time initialized.

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
+++ b/serde-api/src/main/java/io/micronaut/serde/config/annotation/SerdeConfig.java
@@ -16,6 +16,7 @@
 package io.micronaut.serde.config.annotation;
 
 import io.micronaut.context.annotation.Executable;
+import io.micronaut.core.annotation.BuildTimeInit;
 import io.micronaut.core.annotation.Internal;
 
 import java.lang.annotation.ElementType;
@@ -348,6 +349,7 @@ public @interface SerdeConfig {
         /**
          * The discriminator value kind.
          */
+        @BuildTimeInit("io.micronaut.serde.config.annotation.SerdeConfig$SerSubtyped$DiscriminatorValueKind")
         enum DiscriminatorValueKind {
             CLASS_NAME, CLASS_SIMPLE_NAME, NAME, MINIMAL_CLASS
         }


dstepanov:
We not storing enums in the metadata anymore. Were do you see it?

andriy-dmytruk:
> We not storing enums in the metadata anymore. Were do you see it?

I see this:
```java
The following detailed trace displays from which field in the code the object was reached.
Object was reached by
  indexing into array java.lang.Object[]@73ee4e61: [Ljava.lang.Object;@73ee4e61 at index 5
  reading field java.util.ImmutableCollections$MapN.table of constant 
    java.util.ImmutableCollections$MapN@4866fd5f: {dp=@class, dv=CLASS_NAME, dt=PROPERTY, discriminatorVisible=false}
  indexing into array java.lang.Object[]@64f19602: [Ljava.lang.Object;@64f19602 at index 7
  reading field java.util.ImmutableCollections$MapN.table of constant 
    java.util.ImmutableCollections$MapN@50379878: {com.fasterxml.jackson.annotation.JsonRootName={value=dataCenterInfo}, io.micron...
  reading field io.micronaut.inject.annotation.DefaultAnnotationMetadata.allAnnotations of constant 
    io.micronaut.inject.annotation.DefaultAnnotationMetadata@635b6c9c: io.micronaut.inject.annotation.DefaultAnnotationMetadata@635b6c9c
  scanning root io.micronaut.inject.annotation.DefaultAnnotationMetadata@635b6c9c: io.micronaut.inject.annotation.DefaultAnnotationMetadata@635b6c9c embedded in
    io.micronaut.discovery.eureka.client.v2.$DataCenterInfo$Introspection.<init>(Unknown Source)
  parsing method io.micronaut.discovery.eureka.client.v2.$DataCenterInfo$Introspection.<init>(Unknown Source) reachable via the parsing context <no parsing context available> 

```

It has `"io.micronaut.serde.config.annotation.SerdeConfig$SerSubtyped", Map.of("discriminatorVisible", false, "dp", "@class", "dt", "PROPERTY", "dv", DiscriminatorValueKind.CLASS_NAME))` in the `micronaut-discovery-client:4.4.0`

dstepanov:
I don't think that version has the fix.
https://github.com/micronaut-projects/micronaut-core/pull/11160

andriy-dmytruk:
Yes, you are right. The fix is only in `4.6.5` while that version uses `4.6.1`. Thanks!

PR #1011, opened 2025-01-09
Remove use of AliasFor and use mappers for Serde annotations

graemerocher:
using `@AliasFor` adds annotations as stereotypes which causes problems right now when you have multiple annotations as the values from the stereotypes are not resolved. This resolves that by using mappers instead


dstepanov (approved):

graemerocher:
Fixed in https://github.com/micronaut-projects/micronaut-serialization/pull/1012

PR #818, opened 2024-04-11
Improve Enum handling parity with Jackson

oujesky:
Improve enum serialization and deserialization to move closer to parity with Jackson

* Additional Jackson annotations support:
  * deserialize enums with `@JsonValue`
  * serialize and deserialize enums with `@JsonProperty`
  * This follows the same precedence in evaluation as Jackson:
     * serialize: @JsonValue > @JsonProperty
     * deserialize: @JsonCreator > @JsonValue > @JsonProperty

* Improved `EnumSet` de/serialization of contained enums

* Added support for `EnumMap`

PR #1019, opened 2025-01-15
Backport enum support

andriy-dmytruk:
Improve Enum handling parity with Jackson (#818)


graemerocher (approved):

andriy-dmytruk:
Since `DeserializableEnumIntrospection` is not available, I had to use reflection `EnumSet.allOf(...)`.
Not sure what the alternative would be.

PR #1030, opened 2025-01-27
Favor record constructor over builder for records

### LINKED ISSUE #11538: Regression bug: nested-record DTO using Lombok
### Expected Behavior

When using nested records as DTO, and:

- Each is annotated with Lombok's `@Builder`
- Each is annotated with validation annotations
- If the HTTP payload is null or empty, every field (including nested-record fields) should be validated.
- If the HTTP payload is valid, the DTO should be deserialized correctly

With the following versions, it works well:

- micronaut-parent 4.6.2
- micronaut.version set to 4.6.5

### Actual Behaviour

1. If the payload is null or empty, nested-record fields aren't validated, and the nested-record is shown as null:

### Output example using a newer Micronaut version (bugged)

```
jakarta.validation.ConstraintViolationException:
get.<return value>.id: must not be null,
get.<return value>.commonDTO: must not be null  // should list validations for each field instead
```

### Output example using a previous Micronaut version (not bugged)

```
jakarta.validation.ConstraintViolationException:
get.<return value>.id: must not be null,
get.<return value>.commonDTO.commonField: must not be blank,
get.<return value>.commonDTO.commonField2: must not be null
```

2. If the payload is valid, the DTO fails to be deserialized

### Output example using a newer Micronaut version (bugged)

```
Expecting code not to raise a throwable but caught
"jakarta.validation.ConstraintViolationException: get.<return value>.commonDTO: must not be null
```

### Output example using a previous Micronaut version (not bugged)

```
Test passes without throwing and exception
```

This happens at least with the following versions (not sure when the bug was introduced):

- micronaut-parent 4.7.4
- micronaut.version set to 4.7.4

### Steps To Reproduce

I attached an example app with everything set up for you, but if you want to do it manually:

1. Create a new Micronaut project with HTTP client, Lombok and WireMock (or a similar one)
2. Create 2 DTOs and make sure one of them contains the other. Annotate the parameter with `@JsonWrapped`, `@NotNull` and `@Valid`
3. Annotate the nested one's fields with validation annotations (e.g., `@NotNull`, `@NotBlank`)
4. Create an HTTP client with a method that has the top-level DTO as return type
5. Create a test, use WireMock (or a similar one) to stub a null/empty HTTP response and write a simple assert (e.g., assert that call shouldn't throw). When you run it, you'll get an exception saying that the nested record is null (at this point is not validating every field as it should).
6. Create a test, use WireMock (or a similar one) to stub a valid HTTP response and write a simple assert (e.g., assert that call shouldn't throw). When you run it, you'll get an exception saying that the nested record is null, even though the payload is valid, and it shouldn't be deserialized as null.
7. Now, change the Micronaut parent and Micronaut Core versions to 4.6.2 and 4.6.5, respectively.
8. Clean-compile and re-run the tests, and this time they will pass.

I tried setting `<micronaut.core.version>4.7.12</micronaut.core.version>`, but the bug keeps happening.

Let me know if more information is required. Thanks!

### Environment Information

- OS: Linux / Windows / MacOS
- JDK Version: 21

### Example Application

https://github.com/Nahuel92/poc-projects/tree/master/micronaut/micronaut-deserialization-lombok-bug

### Version

4.7.4

graemerocher:
With records the record components define the annotations necessary for deserialization so deserialisation should ignore the builder. Fixes https://github.com/micronaut-projects/micronaut-core/issues/11538 which is a regression.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,6 +14,7 @@ oracle-jdbc-driver = "23.6.0.24.10"
 jetbrains-annotations = "24.1.0"
 jmh = "1.37"
 groovy = "4.0.22"
+lombok = "1.18.36"
 
 micronaut = "4.7.12"
 micronaut-platform = "4.7.4"
@@ -48,6 +49,8 @@ jetbrains-annotations = { module = "org.jetbrains:annotations", version.ref = "j
 aws-lambda-events = { module = "com.amazonaws:aws-lambda-java-events", version.ref = "aws-lambda-events" }
 aws-lambda-serialization = { module = "com.amazonaws:aws-lambda-java-serialization", version.ref = "aws-lambda-serialization" }
 graal-svm = { module = "org.graalvm.sdk:nativeimage", version.ref = "graal-svm" }
+lombok = { module = "org.projectlombok:lombok", version.ref = "lombok" }
+
 microstream-storage-restclient = { module = "one.microstream:microstream-storage-restclient", version.ref = "microstream-storage-restclient" }
 junit-jupiter-engine = { module = "org.junit.jupiter:junit-jupiter-engine" }
 junit-jupiter-api = { module = "org.junit.jupiter:junit-jupiter-api" }
--- a/serde-jackson/build.gradle.kts
+++ b/serde-jackson/build.gradle.kts
@@ -17,6 +17,8 @@ dependencies {
     testAnnotationProcessor(projects.micronautSerdeProcessor)
 
     testCompileOnly(mn.micronaut.inject.groovy)
+    testImplementation(libs.lombok)
+    testAnnotationProcessor(libs.lombok)
 
     testImplementation(projects.micronautSerdeProcessor)
     testImplementation(projects.micronautSerdeTck)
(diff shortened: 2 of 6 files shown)


gradle/libs.versions.toml
```diff
@@ -48,6 +49,8 @@ jetbrains-annotations = { module = "org.jetbrains:annotations", version.ref = "j
 aws-lambda-events = { module = "com.amazonaws:aws-lambda-java-events", version.ref = "aws-lambda-events" }
 aws-lambda-serialization = { module = "com.amazonaws:aws-lambda-java-serialization", version.ref = "aws-lambda-serialization" }
 graal-svm = { module = "org.graalvm.sdk:nativeimage", version.ref = "graal-svm" }
+lombok = { module = "org.projectlombok:lombok", version.ref = "lombok" }
```
yawkat, line 52: really want to open that can of worms? Maybe just write the builder manually?

yawkat (approved):

gradle/libs.versions.toml
graemerocher: not really but I don't know another way to test

PR #1035, opened 2025-02-06
serde to Core 4.8.0

sdelamo:


dstepanov:
https://github.com/micronaut-projects/micronaut-serialization/pull/1035 + looks like Jackson implemented some of the features that are missing vs serde, so the test should be moved to the common TCK

PR #1050, opened 2025-02-24
Implement `@JsonInclude(content=`

dstepanov:


serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedMapSerializer.java
```diff
@@ -100,37 +162,80 @@ public void serialize(Encoder encoder, EncoderContext context, Argument<? extend
...
+                                    case NON_NULL:
+                                        if (v != null) {
+                                            return false;
+                                        }
+                                        break;
+                                    default:
+                                        return false;
+                                }
+                            } catch (SerdeException e) {
+                                throw new RuntimeException(e);
```
graemerocher: need to throw a more specific exception here

test-suite-tck-jackson-databind/src/test/groovy/io/micronaut/serde/tck/jackson/databind/DatabindJsonGetterSetterSpec.groovy
```diff
@@ -82,53 +82,4 @@ class Test {
             context.close()
     }
 
-    void "test json any getter / setter - records"() {
```
graemerocher, line 85: I assume these deleted tests were moved to the TCK?

graemerocher (changes requested):

serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedMapSerializer.java
dstepanov: Fixed

test-suite-tck-jackson-databind/src/test/groovy/io/micronaut/serde/tck/jackson/databind/DatabindJsonGetterSetterSpec.groovy
dstepanov: Yes

serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedMapSerializer.java
graemerocher: What I mean was a more specific exception type. If this is a configuration error maybe `ConfigurationException`?

dstepanov: I have added a sneaky throw so the original exception is thrown withou changes

serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedObjectSerializer.java
```diff
@@ -103,6 +104,22 @@ public void serializeInto(Encoder encoder, EncoderContext context, Argument<? ex
                     default:
                         // fall through
                 }
+                if (property.include != SerdeConfig.SerInclude.ALWAYS && property.include != SerdeConfig.SerInclude.NON_NULL) {
+                    switch (property.includeContent) {
```
yawkat: I don't think jackson does it this way

yawkat: I think it's only maps: https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-annotations/2.18.2/com/fasterxml/jackson/annotation/JsonInclude.html#content--

serde-support/src/main/java/io/micronaut/serde/support/serdes/StringSerde.java
```diff
@@ -48,6 +48,11 @@ public String deserializeNullable(@NonNull Decoder decoder, @NonNull DecoderCont
         return decoder.decodeStringNullable();
     }
 
+    @Override
+    public boolean isAbsent(EncoderContext context, String value) {
+        return value == null;
```
yawkat: From the JsonInclude javadoc, I think this should be the default impl of isAbsent for all types: https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-annotations/latest/com/fasterxml/jackson/annotation/JsonInclude.Include.html#NON_ABSENT

serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedObjectSerializer.java
dstepanov: Optional too

serde-support/src/main/java/io/micronaut/serde/support/serdes/StringSerde.java
dstepanov: Yeah it's the same as the super method

serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedObjectSerializer.java
dstepanov: Removed the common methods and logic and moved it to the optional

dstepanov: It's a bit strange that you cannot exclude empty values from List

yawkat (approved):

graemerocher (approved):

PR #1083, opened 2025-04-16
Align `@JsonTypeInfo(defaultImpl)`, support subtype deduction for unwrapped

dstepanov:
.


serde-api/src/main/java/io/micronaut/serde/config/DeserializationConfiguration.java
```diff
@@ -52,4 +53,14 @@ public interface DeserializationConfiguration {
...
         return false;
     }
+
+    /**
+     * Whether the supertype is used by default when no supertype is resolved.
+     * @return True to avoid the supertype and use `defaultImpl` property
+     */
+    @NextMajorVersion("Inline to true to have the behaviour the same as for Jackson")
+    @Bindable(defaultValue = StringUtils.FALSE)
+    default boolean isAvoidSupertypeSubtype() {
```
graemerocher: This method name is a bit confusing and so is the description. Can we rename it to `subtypes-require-default-impl: true` which I think is more descriptive of what it is?

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBeanSubtypeInfo.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        final Class<T> superType = introspection.getBeanType();
+        final Collection<BeanIntrospection<? extends T>> subtypeIntrospections =
+            decoderContext.getDeserializableSubtypes(superType);
+        Map<String, DeserBean<? extends T>> subtypes = CollectionUtils.newHashMap(subtypeIntrospections.size());
+        Class<? extends T> defaultType = introspection.classValue(DefaultImplementation.class)
+            .orElseGet(() -> introspection.classValue(SerdeConfig.SerSubtyped.DEFAULT_IMPL).orElse(null));
+        if (defaultType == null && !deserializationConfiguration.isAvoidSupertypeSubtype()) {
+            // Jackson always requires to use `defaultImpl` but our initial implementation always takes the supertype
```
graemerocher, line 102: should there be some kind of failure if defaultImpl is not specified?

graemerocher (changes requested):

serde-support/src/main/java/io/micronaut/serde/support/deserializers/DeserBeanSubtypeInfo.java
dstepanov: It will fail at runtime when the subtypes fails to be resolved.

graemerocher (approved):

PR #1096, opened 2025-05-21
Tracking of the referenced path for exceptions

### LINKED ISSUE #553: Ability to customize error messages
### Feature description

Currently we obtain something like:

```
Unable to decode request body: Error decoding JSON stream for type [res]: Error decoding property [<some class> property1] of type [class <some class>]: Error decoding property [<some other class> property1] of type [class <some other class>]: Unexpected token VALUE_STRING, expected START_OBJECT
```

We wanted to be able to customize something like:

```
Invalid json for <fieldName> for class <class name>
```

### LINKED ISSUE #609: No context for unknown property and invalid enum value errors
### Expected Behavior

1. Missing context for errors

I expect the `SerdeException` (or one of it's subclasses, `InvalidFormatException`, etc) to contain some context about the location of an error it encounters while trying to deserialize.

Jackson provides a `getPath()` method in it's `JsonMappingException` which is the super of things like `InvalidFormatException` and `InvalidPropertyException`, this allows one access to the json path in question.

micronaut-serialization should provide similar functionality. 

At the moment it's not possible to write an `ExceptionHandler` to handle these situation and return a custom error message with the details.

2. Incorrect generic exception thrown 

**Invalid enum values**

A `SerdeException` is thrown instead of it's subclass `InvalidFormatException`, 

Minified stacktrace:
```
io.micronaut.core.convert.exceptions.ConversionErrorException: Failed to convert argument [test] for value [null] due to: Error decoding property [TestEnum enum] of type [class mktplc.error.JacksonExceptionHandlerTest$Test]: No enum constant mktplc.error.JacksonExceptionHandlerTest.TestEnum.invalid
....
Caused by: io.micronaut.serde.exceptions.SerdeException: Error decoding property [TestEnum enum] of type [class mktplc.error.JacksonExceptionHandlerTest$Test]: No enum constant mktplc.error.JacksonExceptionHandlerTest.TestEnum.invalid
....
Caused by: java.lang.IllegalArgumentException: No enum constant mktplc.error.JacksonExceptionHandlerTest.TestEnum.invalid
```

If `InvalidFormatException` was thrown we would at least have access to it's `originalValue`.

**Unknown property (when `ignore-unknown` is set to `false`)**

When an unknown property is encountered Jackson has an `IgnoredPropertyException` to represent this, along with appropriate context about the field/property in question.

micronaut-serialization should provide similar functionality. 

Minified stacktrace:
```
io.micronaut.core.convert.exceptions.ConversionErrorException: Failed to convert argument [test] for value [null] due to: Unknown property [unknown] encountered during deserialization of type: Test test
...
Caused by: io.micronaut.serde.exceptions.SerdeException: Unknown property [unknown] encountered during deserialization of type: Test test
```

### Actual Behaviour

See Expected Behavior

### Steps To Reproduce

```kotlin
    @Controller("/test")
    class TestController {
        @Post
        @Consumes(MediaType.APPLICATION_JSON)
        fun jackson(@Body test: Test) {
            // no-op
            LOG.info { "Request Body: $test" }
        }
    }
    
    @Serdeable
    class Test {
        var string: String? = null
        var enum: TestEnum? = null
    }
    
    @Serdeable
    enum class TestEnum { A, B, C }
    
    // apologies for this not being immediately compileable 
    
    // invalid enum value
    val ignore: HttpResponse<Void> = httpClient.toBlocking()
    .exchange(
        HttpRequest.POST(embeddedServer.uri.toASCIIString() + "/test", "{\"enum\":\"invalid\"}")
            .contentType(MediaType.APPLICATION_JSON_TYPE)
    )
    
    // unknown property 
    val ignore: HttpResponse<Void> = httpClient.toBlocking()
        .exchange(
            HttpRequest.POST(embeddedServer.uri.toASCIIString() + "/test", "{\"unknown\":\"invalid\"}")
                .contentType(MediaType.APPLICATION_JSON_TYPE)
        )
```

### Environment Information

MacOs
JDK 17
Micronaut Platform 4.1.5

### Example Application

_No response_

### Version

Micronaut Platform 4.1.5

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/609
Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/553

### DIFF
--- a/serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedIterableSerializer.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedIterableSerializer.java
@@ -19,6 +19,8 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.type.Argument;
 import io.micronaut.serde.Encoder;
 import io.micronaut.serde.Serializer;
+import io.micronaut.serde.exceptions.SerdeException;
+import io.micronaut.serde.exceptions.path.ReferencePath;
 
 import java.io.IOException;
 import java.util.Collection;
@@ -44,13 +46,20 @@ final class CustomizedIterableSerializer<T> implements Serializer<Iterable<T>> {
     @Override
     public void serialize(Encoder encoder, EncoderContext context, Argument<? extends Iterable<T>> type, Iterable<T> value)
         throws IOException {
+        int index = 0;
         try (Encoder array = encoder.encodeArray(type)) {
-            for (T t : value) {
-                if (t == null) {
-                    array.encodeNull();
-                } else {
-                    componentSerializer.serialize(array, context, generic, t);
+            try {
+                for (T t : value) {
+                    if (t == null) {
+                        array.encodeNull();
+                    } else {
+                        componentSerializer.serialize(array, context, generic, t);
+                    }
                 }
+                index++;
+            } catch (SerdeException e) {
+                e.getPath().add(ReferencePath.ofCollection(value.getClass(), type, index));
+                throw e;
             }
         }
     }
--- a/serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedIterableSerializer.java
+++ b/serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedIterableSerializer.java
@@ -48,18 +48,18 @@ final class CustomizedIterableSerializer<T> implements Serializer<Iterable<T>> {
         throws IOException {
         int index = 0;
         try (Encoder array = encoder.encodeArray(type)) {
-            try {
-                for (T t : value) {
+            for (T t : value) {
+                try {
                     if (t == null) {
                         array.encodeNull();
                     } else {
                         componentSerializer.serialize(array, context, generic, t);
                     }
+                    index++;
+                } catch (SerdeException e) {
+                    e.getPath().add(ReferencePath.ofCollection(value.getClass(), type, index));
+                    throw e;
                 }
-                index++;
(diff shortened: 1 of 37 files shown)


serde-support/src/main/java/io/micronaut/serde/support/serializers/CustomizedIterableSerializer.java
```diff
@@ -44,13 +46,20 @@ final class CustomizedIterableSerializer<T> implements Serializer<Iterable<T>> {
...
-                    componentSerializer.serialize(array, context, generic, t);
+            try {
+                for (T t : value) {
+                    if (t == null) {
+                        array.encodeNull();
+                    } else {
+                        componentSerializer.serialize(array, context, generic, t);
+                    }
                 }
+                index++;
```
yawkat: should be inside the loop

serde-support/src/main/java/io/micronaut/serde/support/serializers/StreamSerializer.java
```diff
@@ -47,12 +48,18 @@ public void serialize(Encoder encoder, EncoderContext context, Argument<? extend
...
-                        );
+                    try {
+                        componentSerializer
+                            .serialize(
+                                arrayEncoder,
+                                context, generic, itr.next()
+                            );
+                        index++;
+                    } catch (SerdeException e) {
+                        e.getPath().add(ReferencePath.ofCollection(value.getClass(), type, index));
```
yawkat, line 61: rethrow!

yawkat (approved):

graemerocher (approved):

PR #1097, opened 2025-05-22
Support `@JsonInclude(NON_DEFAULT)`

dstepanov:
Except this scenario:

> When used for a POJO, definition is that only values that differ from the default values of POJO properties are included. This is done by creating an instance of POJO using zero-argument constructor, and accessing property values: value is used as the default value by using equals() method, except for the case where property has `null` value in which case straight null check is used.


graemerocher:
tests are failing

dstepanov:
@graemerocher Fixed

graemerocher (approved):

yawkat (approved):

PR #1099, opened 2025-05-27
JsonUnwrapped not deserialized properly when used as constructor para…

radovanradic:
…meter

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/SerdeJsonUnwrappedSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/SerdeJsonUnwrappedSpec.groovy
@@ -108,6 +108,7 @@ class B {
         def ctx = buildContext("""
 package unwrapped;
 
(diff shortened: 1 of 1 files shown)


serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/SerdeJsonUnwrappedSpec.groovy
```diff
@@ -194,6 +195,12 @@ class Address {
 
     private String street;
 
+    @JsonCreator
+    public Address(CityData cityData, String street) {
```
radovanradic, line 199: Added explicit `@JsonCreator` constructor and then `cityData` is not getting deserialized properly.
If constructor is changed to (explicitly adding `@JsonUnwrapped` to the parameter same as defined on the field)
```
@JsonCreator
    public Address(@JsonUnwrapped(prefix = "cd_") CityData cityData, String street) {
        this.cityData = cityData;
        this.street = street;
    }
```
then deserialization works as expected.

PR #1168, opened 2025-09-26
Record deserialization fails when there are custom constructors

radovanradic:
It fails when `micronaut.serde.deserialization.ignore-unknown=false` and throws `unknown property theDate` and when ignore-unknown=true it does not fail but picks wrong constructor and does not deserialize as expected


serde-jackson/src/test/groovy/io/micronaut/serde/jackson/MyRecord.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.serde.annotation.Serdeable;
+
+import java.time.Instant;
+import java.util.Date;
+
+@Serdeable
+record MyRecord(String fooBar, int abcXyz, @Nullable Date theDate, @Nullable String otherStr) {
+
```
radovanradic, line 12: Adding `@JsonCreator` here resolves the issue btw.

dstepanov:
I think this needs to be fixed in core

graemerocher:
@dstepanov can you look at it? Seems like a bug to me

PR #1174, opened 2025-10-06
Fix deduction with custom names defined

dstepanov:


graemerocher (approved):

graemerocher:
Seems there are test failures

PR #1169, opened 2025-09-29
Correct @JsonInclude for @JsonUnwrapped

radovanradic:

PR #1196, opened 2025-11-06
Fix subtypes being set to null for empty JSON

andriy-dmytruk:
Given the following classes:

```java
@Serdeable
@JsonSubTypes({
    @JsonSubTypes.Type(value = ChildA.class, name = "subtype-a"),
    @JsonSubTypes.Type(value = ChildB.class, name = "subtype-b")
})
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "subtype")
sealed interface Parent permits ChildA, ChildB {
}

@Serdeable
final class ChildA implements Parent {
    private String propA;
    public ChildA(String propA) { this.propA = propA; }
    public String getPropA() { return propA; }
}
```

Deserialization would set the value to null when the JSON is of form `{"subtype":"subtype-a"}`. This is incorrect, as information is lost (the subtype information in this case).

@radovanradic Please take a look. It seems like this reverts some of the changes you made.

### DIFF
--- a/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/JacksonBasicSerdeSpec.groovy
+++ b/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/JacksonBasicSerdeSpec.groovy
@@ -35,8 +35,9 @@ class JacksonBasicSerdeSpec extends AbstractBasicSerdeSpec {
         obj = new MyBeanWithNestedObject("id2", null)
         json = jsonMapper.writeValueAsString(obj)
         result = jsonMapper.readValue(json, MyBeanWithNestedObject)
-        then:
-        obj == result
+        then: "The result is different, as always has added the type"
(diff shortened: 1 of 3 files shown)


radovanradic:
It might break tests fixed in this PR https://github.com/micronaut-projects/micronaut-serialization/pull/1169
Commits adding tests that were breaking
https://github.com/micronaut-projects/micronaut-serialization/pull/1169/commits/a200d3daeba2bfb967543b6dfcbc4a4846c69f3c
https://github.com/micronaut-projects/micronaut-serialization/pull/1169/commits/d88d7ea8687284e70631da07784b922ad6a8a286
which Denis actually fixed in that PR

andriy-dmytruk:
I don't think `@JsonSubTypes` with `@JsonUnwrapped` will work correctly, but that would be very complex anyway.

graemerocher (approved):

dstepanov:
Next time pls add the test to the tck so we check that Jackson behaves the same. The correct fix might be to see if there are wrapped prperties present

PR #1209, opened 2025-11-29
Fix Test Discovery Failure

lissam1:
This pr fix test discovery failure [see](https://ge.micronaut.io/s/i4i74yimxjuik/failures#1), in the past it was a similar issue in another module and @melix manage to fix the issue, tests pass but not sure if the fix accurate.


melix (changes requested):
The dependencies on `junit.platform.suite` are highly suspicious. I doubt we need these. Probably only `launcher` and `engine`, not more.

lissam1:
> The dependencies on `junit.platform.suite` are highly suspicious. I doubt we need these. Probably only `launcher` and `engine`, not more.

but couple of tests depends on suite `import org.junit.platform.suite.api.Suite;`
e.g **SerializationJacksonDatabindSuite**

PR #1250, opened 2026-02-24
Show example of using micronaut-serialization's ObjectMapper + Jackson without micronaut itself.

### LINKED ISSUE #1241: Using micronaut-serialization standalone, without micronaut itself
I wanted to just create a discussion on this, because using micronaut-serialization without micronaut itself has made a really massive difference to my graalvm native workflow (because i don't have to bother with reflect config.).

Micronaut-serialization can be made to work standalone quite easily, but it is not documented at all how to do this. Considering it would likely be helpful in fixing other people's graalvm issues, i wonder if it's worth having either this documented or suggesting it as an actual replacement for where people would typically use jackson.

jakegt1:
I ended up using this in a personal project that where the main output was a GraalVM native binary. I have had awful amounts of pain with gson or standard jackson with GraalVM, whereas this works perfectly without any requirement to deal with horrible reflection config.

I think this has great benefits for non-web framework users of GraalVM, so thought an example was a good idea. Not sure where else i should be doccing this though..

Closes #1241


gradle/libs.versions.toml
```diff
@@ -21,6 +21,7 @@ micronaut-oracle-cloud = "5.6.2"
 micronaut-reactor = "3.9.1"
 micronaut-validation = "4.12.0"
 micronaut-gradle-plugin = "4.6.2"
+micronaut-inject = "4.10.15"
```
graemerocher: this hard coded version shouldn't be necessary

jakegt1: Changed, thanks.

PR #1283, opened 2026-04-10
feat : Adding support for YAML serialization/deserialization

ZARAYACH:
This pr aims to add support for YAML serialization/deserialization


serde-yaml/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+dependencies {
+    annotationProcessor(projects.micronautSerdeProcessor)
+
+    api(mn.micronaut.context)
+    api(projects.micronautSerdeApi)
+
+    implementation(mn.micronaut.context)
+    implementation(mn.micronaut.core)
+    implementation(projects.micronautSerdeApi)
+    implementation(mn.snakeyaml)
```
graemerocher: I thought there was a lower level dependency we could depend on here?

khamzatMurj: If you meant low level one is snakeyaml engine. **No we can't depend on snake-yaml engine**, Bacause The Core BOM currently **exposes only** `org.yaml:snakeyaml`.

 and if you want to introduce `org.snakeyaml:snakeyaml-engine` in the Bom or **libs-version.toml** we only need to replace `org.yaml.snakeyaml.*` APIs with `org.snakeyaml.engine.v2.*` imports with small changes. 

-I've test it locally no break changes on the pr.

khamzatMurj: @graemerocher also the **jackson-databind** itself rely on the **snakeyaml-engine** as dependency as well.
which is technically the same low level classes(Emitter, Event, Node ...) I used in encoder and decoder.

graemerocher: the point is to depend only on org.snakeyaml:snakeyaml-engine not on the entirety of snakeyaml

khamzatMurj: [Introduced](https://github.com/micronaut-projects/micronaut-serialization/pull/1283/commits/aeb8c89323fc93fb3d64487f597d4482bb2801dc) by @sdelamo thank you.

khamzatMurj:
This current implementation now lacking these configuration as the upstream [jackson-dataformat-yaml](https://github.com/FasterXML/jackson-dataformats-text/blob/3.x/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLWriteFeature.java) does : 
- USE_NATIVE_OBJECT_ID
- USE_NATIVE_TYPE_ID

these limitation is because micronaut-serialization not supporting **@JsonIdentityInfo** & **@JsonIdentityReference**. imo Native type IDs would need a separate cross-module feature.

these are the jackson tests for the desired behavior : [Tests](https://github.com/FasterXML/jackson-dataformats-text/blob/3.x/yaml/src/test/java/tools/jackson/dataformat/yaml/misc/ObjectIdTest.java).

PR #1307, opened 2026-05-04
Improve and add more tests for compile-time serdes

dstepanov:
Key changes:

  - Added @SerdeableGenerated in serde-api/src/main/java/io/micronaut/serde/annotation/SerdeableGenerated.java with required, skip, skipSerializer, and skipDeserializer.
  - Added processor validation so required=true fails compilation when a non-skipped generated serializer/deserializer is unsupported.
  - Updated sourcegen naming and serializer/deserializer generation: indexed fields/locals, encodeObject plus serializeInto, static key fields for property paths, string switch for more than 5 properties, primitive boolean
    decode without decodeBooleanNullable, and pre-constructor missing-property checks.
  - Added compile-time fixture tests under io.micronaut.serde.jackson.compiletime.
  - Annotated eligible benchmark data/sourcegen benchmark shapes directly with @SerdeableGenerated; no BenchmarkCompileTimeSerdeTypes fixture is generated/kept.

### DIFF
--- a/serde-api/src/main/java/io/micronaut/serde/annotation/SerdeableGenerated.java
+++ b/serde-api/src/main/java/io/micronaut/serde/annotation/SerdeableGenerated.java
@@ -23,7 +23,22 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * A {@link Serdeable} type that opts into compile-time generated serializers and deserializers.
+ * Declares the compile-time generated serde contract for a {@link Serdeable} type.
+ *
+ * <p>Micronaut Serialization already attempts to generate compile-time serializers and
+ * deserializers for simple eligible {@link Serdeable} shapes without this annotation. A shape is
+ * currently eligible when it is a simple bean, record, or enum whose serialization model can be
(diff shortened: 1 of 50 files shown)


serde-api/src/main/java/io/micronaut/serde/annotation/SerdeableGenerated.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.micronaut.core.annotation.Experimental;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A {@link Serdeable} type that opts into compile-time generated serializers and deserializers.
```
graemerocher: it should be made clear that compile-time generated serializers are generated by default where possible, so this is not opt-in at the moment

we may want to use this javadoc to also describe when compile time generated serializers/deserializers are generated today

graemerocher (changes requested):

graemerocher (approved):

PR #1318, opened 2026-05-11
Fix nullability and compile-time serdes

dstepanov:
- correct that deserialize method is not nullable vs deserializeNullable
- improve some cases where try-catch was used in loops instead of outside of it
- add more coverage for compile-time serdes
- fixed data zone issue


serde-api/src/main/java/io/micronaut/serde/util/NullableSerde.java
```diff
@@ -1,25 +0,0 @@
...
- */
-package io.micronaut.serde.util;
-
-import io.micronaut.serde.Serde;
-
-/**
- * Specialization of {@link io.micronaut.serde.Serde} for types that allow a value of {@code null}.
- * @param <T> Tge generic type
- */
-public interface NullableSerde<T> extends Serde<T>, NullableDeserializer<T> {
```
graemerocher, line 24: this is likely quite a large breaking change since these nullability types are used often, was this necessary?

graemerocher (changes requested):

graemerocher (approved):

serde-api/src/main/java/io/micronaut/serde/util/NullableSerde.java
dstepanov: Reverted

PR #1331, opened 2026-06-02
JSON-B and JSON-P Spec Compliance

graemerocher:
## Summary

Adds first-class Jakarta JSON-P and JSON-B provider support backed by Micronaut Serialization, with dedicated TCK coverage and internal refactoring needed for spec-compatible JSON-B behavior.

This branch introduces new `serde-jsonp-impl` and `serde-jsonb` modules, wires them into the build, and adds service loader providers for `jakarta.json.spi.JsonProvider` and `jakarta.json.bind.spi.JsonbProvider`. The implementation keeps conversion on Micronaut Serialization codecs where possible, while adding JSON-B-specific runtime metadata support for cases that require JSON-B reflection semantics.

### JSON-P provider

- Adds `serde-jsonp-impl` with `MicronautJsonProvider` as a Jakarta JSON-P provider implementation.
- Implements JSON-P value, builder, parser, generator, pointer, patch, merge patch, reader, writer, and factory behavior on top of Micronaut JSON tree and Serde infrastructure.
- Registers the provider through `META-INF/services/jakarta.json.spi.JsonProvider`.
- Adds JSON-P provider tests and quick-start documentation updates.

### JSON-B provider

- Adds `serde-jsonb` with `MicronautJsonbProvider`, `JsonbFactory`, JSON-B config handling, and service registration through `META-INF/services/jakarta.json.bind.spi.JsonbProvider`.
- Supports generated Micronaut Serialization paths as the default fast path for JSON-B read/write.
- Adds JSON-B-specific encoder and decoder wrappers for spec differences such as binary data strategies, arbitrary number handling, JSON-B temporal behavior, and JSON-P value mapping.
- Adds JSON-B serializers/deserializers for missing scalar or spec-sensitive types, including calendar, `GregorianCalendar`, `OffsetTime`, `ZoneOffset`, `PriorityQueue`, `Optional[]`, and JSON-P values.
- Adds configured adapter, serializer, and deserializer bridges for JSON-B config and annotation-driven customization.

### JSON-B reflection fallback and runtime introspections

- Adds runtime JSON-B introspection support for fallback-only cases where generated Serde metadata is not sufficient for JSON-B rules.
- Keeps generated Serde as the normal path and only uses fallback runtime introspection after JSON-B fallback rules require it.
- Routes ordinary fallback bean serialization and deserialization through Micronaut Serialization serializers, deserializers, and bounded encoder/decoder paths instead of maintaining a parallel reflection conversion engine.
- Retains reflection helpers only for metadata discovery, visibility, creator/default-constructor validation, transient handling, property naming, and annotation lookup.
- Adds `JsonbFallbackCodec` and `JsonbJsonpBridge` to centralize bounded tree/stream encode/decode and JSON-P callback bridging.
- Uses direct CDI API dispatch for JSON-B callback lookup when CDI is present, with CDI as a `compileOnly` dependency and runtime presence checks.

### Serde core/support changes

- Adds runtime resolver introspection support so modules can resolve Serde introspections dynamically when a provider needs runtime metadata.
- Extends Serde configuration with limit and date/time formatting behavior needed by the JSON-B provider.
- Propagates configured Serde limits into JSON-B generated and fallback paths instead of using default limits where configuration is available.
- Adds stricter i-json date/time formatting support and adjusts temporal, number, and object serializer/deserializer support used by JSON-B compliance paths.
- Adds processor mappings/transformers for JSON-B type adapters, type serializers/deserializers, type info, nillable handling, and number format behavior.

### TCK and build integration

- Adds standalone Jakarta JSON-P and JSON-B TCK modules under `tests/`.
- Adds known-failure tracking for TCK runs.
- Adds `.github/workflows/json-tck.yml` and `.github/scripts/json-tck-evidence.sh` for CI and evidence collection.
- Adds the new modules to `settings.gradle.kts` and build wiring.
- Keeps TCK-only dependencies grouped separately in the version catalog with comments and without the `managed-*` prefix.
- Skips binary baseline checks only for new modules introduced by this PR using `enabledAfter("3.1.0")`.

## Reviewer Notes

- Public provider modules are new, but the JSON-B fallback and bridge classes are internal/package-private implementation details.
- JSON-B generated paths remain preferred by default. Runtime reflection-style introspection is fallback-only.
- JSON-B custom adapters, serializers, and deserializers now flow through bounded Serde codec bridges and JSON-P parser/generator bridges rather than string buffering or ad hoc map/list rewrites.
- JSON-B callback components are resolved in this order: Micronaut bean, CDI bean if CDI is present, default constructor fallback.
- TCK dependencies are intentionally cataloged as TCK-only dependencies and are not treated as managed runtime dependencies.

## Verification

The following checks were run successfully on this branch:

- `./gradlew check`
- `./gradlew :micronaut-serde-jsonb:assertNoYassonDependency :micronaut-tests:micronaut-jsonb-tck:jakartaJsonbTck jakartaJsonTck`
- `git diff --check`

A prior full `check` run hit a transient failure in `:test-suite-http-server-tck-netty:test`, but the subsequent full `./gradlew check` rerun passed, including that task.


.github/workflows/json-tck.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: "🧪 Run Jakarta JSON-B TCK"
+        id: jsonb
+        continue-on-error: true
+        env:
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_CACHE_USERNAME: ${{ secrets.GRADLE_ENTERPRISE_CACHE_USERNAME }}
+          DEVELOCITY_CACHE_PASSWORD: ${{ secrets.GRADLE_ENTERPRISE_CACHE_PASSWORD }}
+        run: ./gradlew :micronaut-tests:micronaut-jsonb-tck:jakartaJsonbTck --no-daemon
+
```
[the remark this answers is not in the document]
graemerocher: this is not accurate, the task names are correct

serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
```diff
@@ -78,6 +78,50 @@ default boolean isInetAddressAsNumeric() {
...
+    /**
+     * Control whether date/time values are written in JSON-B strict I-JSON temporal form.
+     *
+     * @return Whether date/time values use strict I-JSON temporal formatting
+     * @since 3.1.0
+     */
+    @Bindable(defaultValue = "false")
+    default boolean isWriteDateTimesAsStrictIJson() {
+        return false;
+    }
```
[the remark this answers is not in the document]
graemerocher: this will eventually target 3.1.0 so no

build.gradle.kts
```diff
@@ -8,10 +8,44 @@ repositories {
...
+    group = "verification"
+    description = "Runs the standalone Jakarta JSON-P and JSON-B TCK modules."
+    dependsOn("jakartaJsonpTck", "jakartaJsonbTck")
+}
+
 if (System.getenv("SONAR_TOKEN") != null) {
     configure<SonarExtension> {
         properties {
             property("sonar.exclusions", "**/example/**")
+            property(
```
yawkat, line 33: why these exclusions?

tests/jsonb-tck/build.gradle.kts
```diff
@@ -0,0 +1,302 @@
...
+val jakartaJsonbTckSigtestClasspath = files(
+    sourceSets.test.get().runtimeClasspath,
+    jakartaJsonbTck,
+    jakartaJsonbTckClasses,
+    jakartaJsonbTckJimage.map { it.dir("java.base") }
+)
+
+data class TckFailure(val classname: String, val name: String)
+
+fun parseXml(file: File) = DocumentBuilderFactory.newInstance()
```
yawkat: i think json would be easier here

tests/jsonp-tck/build.gradle.kts
```diff
@@ -0,0 +1,220 @@
...
+val jakartaJsonpTckSigtestClasspath = files(
+    sourceSets.test.get().runtimeClasspath,
+    jakartaJsonpTck,
+    jakartaJsonpTckClasses,
+    jakartaJsonpTckJimage.map { it.dir("java.base") }
+)
+
+data class TckFailure(val classname: String, val name: String)
+
+fun parseXml(file: File) = DocumentBuilderFactory.newInstance()
```
yawkat: maybe use a gradle convention

build.gradle.kts
graemerocher: because they are covered by the TCK

serde-api/src/main/java/io/micronaut/serde/RuntimeResolverSerdeIntrospections.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Resolver-backed introspections wrapper.
+ */
+final class RuntimeResolverSerdeIntrospections implements SerdeIntrospections {
+    private final SerdeIntrospections delegate;
+    private final RuntimeIntrospectionResolver resolver;
+    private final Map<RuntimeIntrospectionCacheKey, Optional<BeanIntrospection<?>>> runtimeIntrospections = new ConcurrentHashMap<>();
```
yawkat, line 40: do we need to watch size of this cache? probably not…

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * Converts a reflection {@link Type} into a Serde {@link Argument}. This
+     * keeps JSON-B runtime adapters/deserializers on the same generic type model
+     * used by generated Serde code.
+     *
+     * @param runtimeType The reflection type
+     * @return The corresponding Serde argument
+     */
+    @SuppressWarnings("java:S1452")
+    static Argument<?> argument(Type runtimeType) {
```
yawkat, line 143: do we not have a similar utility in core? i could swear ive implemented this before

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbCalendarSerde.java
```diff
@@ -0,0 +1,131 @@
...
+        encoder.encodeString(format(value, strictIJson));
+    }
+
+    @Override
+    public Calendar deserialize(Decoder decoder, DecoderContext decoderContext, Argument<? super Calendar> type) throws IOException {
+        return parse(decoder.decodeString());
+    }
+
+    @Override
+    public @Nullable Calendar deserializeNullable(Decoder decoder, DecoderContext context, Argument<? super Calendar> type) throws IOException {
```
yawkat: thats the default impl

serde-api/src/main/java/io/micronaut/serde/RuntimeResolverSerdeIntrospections.java
graemerocher: don't think so, it is fixed by the number of classes in the project. It could leak if someone did some dynamic class compilation and serialisation but that seems niche and if you are doing that you are probably asking for trouble

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
graemerocher: only thing similar is ReflectTypeElement but that is in the processor module. That said this is likely a candidate to upstream to core, but that is likely a separate enhancement that shouldn't delay this PR.

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbDecoder.java
```diff
@@ -0,0 +1,342 @@
...
+                return BigInteger.valueOf(number.longValue());
+            }
+            if (number instanceof Float || number instanceof Double) {
+                return BigDecimal.valueOf(number.doubleValue()).toBigIntegerExact();
+            }
+            return new BigDecimal(number.toString()).toBigIntegerExact();
+        } catch (ArithmeticException | NumberFormatException e) {
+            throw createDeserializationException("Byte array entry is not an integral value: " + number, number);
+        }
+    }
```
yawkat, line 256: this is extremely inefficient

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
```diff
@@ -0,0 +1,114 @@
...
+    @Singleton
+    Jsonb jsonb(JsonbConfig config,
+                BeanContext beanContext,
+                ObjectMapper objectMapper,
+                SerdeIntrospections serdeIntrospections,
+                SerdeConfiguration serdeConfiguration,
+                SerializationConfiguration serializationConfiguration,
+                DeserializationConfiguration deserializationConfiguration,
+                JsonbConfiguration jsonbConfiguration) {
+        if (jsonbConfiguration.isReflectionEnabled() || MicronautJsonbProvider.MicronautJsonb.hasReflectionOnlyFeatures(config)) {
```
yawkat: if there is auto detection, maybe the config should not be a boolean, but instead "on", "off", "auto". That way, there is no risk of accidentally enabling reflection support if you want it off explicitly but use a reflection-only feature. The builder would throw an exception in that case

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFallbackCodec.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    @SuppressWarnings("java:S2095")
+    <T> void writeValue(JsonGenerator generator, Argument<T> argument, @Nullable T value) throws IOException {
+        if (value == null) {
+            generator.writeNull();
+            return;
+        }
+        Serializer.EncoderContext encoderContext = registry.newEncoderContext(null);
+        Serializer<? super T> serializer = findSerializer(encoderContext, argument);
+        serializer.serialize(new JsonbEncoder(JacksonEncoder.create(generator, limits), binaryDataStrategy), encoderContext, argument, value);
```
yawkat, line 133: dependency on jackson?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbDecoder.java
graemerocher: what do you suggest as an alternative?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFallbackCodec.java
graemerocher: sure this is a JSON-B implementation, we choose what we want to use underneath to implement the spec. The goal is to not have to use an existing implementation. What do you suggest otherwise?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
graemerocher: let me try to understand.

- on = you are ok with reflection, not exception
- auto = it is enabled but if reflection paths are hit you want an exception
- off = no reflection, currently this would be an exception anyway same as we have now (no serialisable introspection found)

I am not sure I understand the value to be honest

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbDecoder.java
yawkat: don't have to go through BigDecimal for most number types

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
yawkat: * on: always use MicronautJsonbReflectionProvider
* off: always use MicronautJsonbProvider (throw an exception if hasReflectionOnlyFeatures)
* auto: use provider depending on hasReflectionOnlyFeatures

The current code only supports the equivalent of "on" and "auto", "off" is what I propose as the new setting to avoid accidentally using reflection

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFallbackCodec.java
yawkat: i thought this would use standard serde apis, but okay

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
graemerocher: ok I see

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFallbackCodec.java
graemerocher: ok let me see if it can be made to use standard APIs, not are it can. Something has to encode the JSON otherwise it means duplicating JacksonEncoder into this module

yawkat: i think it probably can't, standard serde API (intentionally) lacks token peeking functionality. just leave it as-is

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbOptionalArraySerde.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public Deserializer<Optional[]> createSpecific(DecoderContext context, Argument<? super Optional[]> type) throws SerdeException {
+        Argument<Optional> componentType = componentType(type);
+        Deserializer<? extends Optional> componentDeserializer = context.findDeserializer(componentType).createSpecific(context, componentType);
+        return new Deserializer<>() {
+            @Override
+            public Optional[] deserialize(Decoder decoder, DecoderContext decoderContext, Argument<? super Optional[]> arrayType) throws IOException {
+                Decoder array = decoder.decodeArray();
+                Optional[] values = new Optional[10];
```
yawkat: use a list and toArray it instead

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbOptionalArraySerde.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.util.Optional;
+
+/**
+ * JSON-B default mapping for arrays of {@link Optional}.
+ */
+@SuppressWarnings({"rawtypes", "unused"})
+@Internal
+@Singleton
+final class JsonbOptionalArraySerde implements Serializer<Optional[]>, Deserializer<Optional[]>, SerdeRegistrar<Optional[]> {
```
yawkat: why is this serde necessary in the first place? it looks like it should be default behavior, maybe a gap in the standard serdes?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbPriorityQueueSerde.java
```diff
@@ -0,0 +1,101 @@
...
+                return deserialize(decoder, context, type);
+            }
+
+            @Override
+            public PriorityQueue<Object> deserialize(Decoder decoder, DecoderContext context, Argument<? super PriorityQueue<Object>> type) throws IOException {
+                PriorityQueue<Object> queue = new PriorityQueue<>();
+                Decoder array = decoder.decodeArray();
+                while (array.hasNextArrayValue()) {
+                    Object item = elementDeserializer.deserializeNullable(array, context, elementType);
+                    if (item != null) {
```
yawkat: it's a bit weird to drop null items, is it necessary for compliance?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbPriorityQueueSerde.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.IOException;
+import java.util.PriorityQueue;
+
+/**
+ * JSON-B default mapping for {@link PriorityQueue}.
+ */
+@Internal
+@Singleton
+final class JsonbPriorityQueueSerde implements SerdeRegistrar<PriorityQueue<Object>> {
```
yawkat: maybe this serde should be standard

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbReflectionUtil.java
```diff
@@ -0,0 +1,484 @@
...
+     * models that do not have a creator constructor.
+     *
+     * @param type The type to instantiate
+     * @param <T> The bean type
+     * @return The new instance
+     */
+    static <T> T instantiate(Class<T> type) {
+        Constructor<T> constructor = ReflectionUtils.findConstructor(type)
+            .orElseThrow(() -> new JsonbException("No default constructor available for JSON-B fallback type " + type.getName()));
+        constructor.setAccessible(true);
```
yawkat: i think this produces log warnings in newer jdks

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbOptionalArraySerde.java
graemerocher: all the added serdes are for spec compliance. The TCK has tests for this case. If we want to be compliant with the spec we need this serde. It is debatable whether this serde should be in the standard ones or not. Frankly an array of Optional seems a weird case to me.

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbReflectionUtil.java
```diff
@@ -0,0 +1,484 @@
...
+     */
+    static boolean isGetter(Method method) {
+        return method.getParameterCount() == 0
+            && !method.getReturnType().equals(Void.TYPE)
+            && !method.isSynthetic()
+            && !method.isBridge()
+            && !method.isAnnotationPresent(JsonbTransient.class)
+            && method.getDeclaringClass() != Object.class
+            && ((method.getName().startsWith("get") && method.getName().length() > 3)
+            || (method.getName().startsWith("is") && method.getName().length() > 2 && method.getReturnType() == boolean.class));
```
yawkat, line 118: partially repeated with isGetterName

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbReflectionUtil.java
graemerocher: probably, but the spec has cases that require reflection, not sure what the alternative is

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbReflectionUtil.java
```diff
@@ -0,0 +1,484 @@
...
+
+    private static boolean isSetterName(Method method) {
+        return method.getName().startsWith("set") && method.getName().length() > 3;
+    }
+
+    private static String splitCamelCase(String name, String separator) {
+        StringBuilder builder = new StringBuilder(name.length() + 8);
+        for (int i = 0; i < name.length(); i++) {
+            char character = name.charAt(i);
+            if (i > 0 && Character.isUpperCase(character)) {
```
yawkat, line 457: no acronym handling?

graemerocher: spec and TCK doesn't require it. Not sure acronyms are in the java bean spec to be honest

yawkat: bean spec is very unclear, had a discussion about that just a while ago ^^ https://github.com/FasterXML/jackson-databind/issues/5712

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbReflectionUtil.java
yawkat: theres a trySetAccessible method ive seen people use which may have more graceful error handling without log statements. but i dont know how it is on newer jdks, let AI figure it out :smile:

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbOptionalArraySerde.java
yawkat: I'm just surprised it doesn't work out of the box, since an array of any object should work, including optional objects

graemerocher: Yeah unclear to me why it doesn't work

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbRuntimeBeanConstructor.java
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * @param executable The constructor or factory method
+     * @return The constructor arguments exposed to Serde
+     */
+    private static Argument<?>[] arguments(Executable executable) {
+        Parameter[] parameters = executable.getParameters();
+        Type[] genericTypes = executable instanceof Constructor<?> constructor
+            ? constructor.getGenericParameterTypes()
+            : executable.getGenericParameterTypes();
+        Argument<?>[] arguments = new Argument<?>[parameters.length];
```
yawkat: why this check?

dstepanov:
Move all the new JSONB serializers / deserializers  / serdes into a new package and have the package condition. Are those discovered by the value or only SERIALIZER_CLASS/DESERIALIZER_CLASS? Maybe we can put `@Bean(exposes=type)`.

.github/workflows/json-tck.yml
```diff
@@ -0,0 +1,71 @@
+name: Jakarta JSON TCK
```
dstepanov, line 1: Why do we need an extra workflow? Pls remove it, this will be overridden anyway

config/checkstyle/suppressions.xml
```diff
@@ -9,4 +9,5 @@
               <!-- files="DefaultBeanContext.java|BeanDefinitionWriter.java|DefaultHttpClient.java"/> -->
 
     <suppress checks="MissingJavadocType" files=".*doc-examples.*" />
+    <suppress checks="FileLength" files="MicronautJsonProvider.java" />
```
dstepanov: Extract all inner classes from MicronautJsonProvider and remove this

serde-api/src/main/java/io/micronaut/serde/config/DeserializationConfiguration.java
```diff
@@ -147,6 +147,17 @@ default boolean adjustDatesToContextTimeZone() {
...
     }
 
+    /**
+     * Determines whether every creator parameter must be present in the JSON input.
+     *
+     * @return {@code true} if all creator parameters are required
+     * @since 3.1.0
+     */
+    @Bindable(defaultValue = StringUtils.FALSE)
+    default boolean isRequireAllCreatorParameters() {
```
dstepanov, line 157: Is it true by default in JSONB? Or why do we need this?

serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
```diff
@@ -78,6 +78,50 @@ default boolean isInetAddressAsNumeric() {
...
     boolean isWriteBinaryAsArray();
 
+    /**
+     * Control whether {@link java.time.Duration} values are written and read as ISO-8601 strings.
+     *
+     * @return Whether durations are represented as strings
+     * @since 3.1.0
+     */
+    @Bindable(defaultValue = "false")
+    default boolean isWriteDurationsAsStrings() {
```
dstepanov, line 88: I wonder if this and isWriteJavaUtilDatesWithZoneId can be replaced with FormattedSerde/FormatConfiguration

serde-api/src/main/java/io/micronaut/serde/ObjectMapper.java
```diff
@@ -64,6 +64,26 @@ default ObjectMapper cloneWithConfiguration(
...
+     * {@code null} configuration parameter indicates the old configuration should be used.
+     *
+     * @param configuration The {@link SerdeConfiguration}
+     * @param serializationConfiguration The {@link SerializationConfiguration}
+     * @param deserializationConfiguration The {@link DeserializationConfiguration}
+     * @param introspections The {@link SerdeIntrospections}
+     * @return A new {@link JsonMapper} with the updated config and introspections
+     * @since 3.1.0
+     */
+    default ObjectMapper cloneWithConfiguration(
```
dstepanov, line 78: Maybe we should move all the options to features and have something like `withFeatures`/`withoutFeatures`

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonDecoder.java
```diff
@@ -1163,6 +1163,44 @@ private boolean endUnwrapArray() throws IOException {
...
     }
 
+    /**
+     * Peeks the next token for integrations that need direct JSON token streaming.
+     *
+     * @return The next token
+     * @throws IOException If an unrecoverable error occurs
+     */
+    @Internal
+    public JsonToken peekTokenForStreaming() throws IOException {
```
dstepanov, line 1173: I don't think it's correct that JSONB is only supported for Jackson, this probably should be some interface that decoders can support

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
dstepanov, line 143: Can we combine this code in Core with io.micronaut.context.AnnotationReflectionUtils? Maybe a new class or something

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        /**
+         * Reflectively instantiates a JSON-B component as the final fallback.
+         *
+         * @param type The component type
+         * @param <T> The component type
+         * @return The component instance
+         */
+        static <T> T instantiate(Class<T> type) {
```
dstepanov, line 207: This would be nice to have in ReflectionUtils

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
```diff
@@ -0,0 +1,253 @@
...
+         * Determines whether the optional CDI API is available. The JSON-B
+         * module compiles against CDI as {@code compileOnly}; this guard keeps
+         * callback lookup usable when applications do not include CDI.
+         *
+         * @return Whether CDI support can be directly called
+         */
+        static boolean isCdiSupportPresent() {
+            try {
+                //noinspection ConstantConditions
+                return CDI.class != null;
```
dstepanov, line 227: Use ClassUtils

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbCalendarSerde.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public Argument<Calendar> getType() {
+        return CALENDAR;
+    }
+
+    @Override
+    public Iterable<Argument<?>> getTypes() {
+        return SerdeRegistrar.super.getTypes();
```
dstepanov: Is it needed?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbConfiguredSerde.java
```diff
@@ -0,0 +1,246 @@
...
+    private static void serializeIterable(Encoder encoder,
+                                          EncoderContext context,
+                                          Argument<?> type,
+                                          Iterable<?> iterable,
+                                          JsonbRuntimeCustomizations customizations,
+                                          Serializer<? super JsonNode> jsonNodeSerializer,
+                                          JsonbFallbackCodec codec) throws Exception {
+        Argument<?> elementType = type.getTypeParameters()[0];
+        Encoder array = encoder.encodeArray(type);
+        for (Object item : iterable) {
```
dstepanov, line 117: The best would be to track the last item class and not resolve everything all the time

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbDecoder.java
```diff
@@ -0,0 +1,349 @@
...
+ * JSON-B decoder wrapper around a Serde decoder.
+ * <p>
+ * The wrapper keeps normal Serde deserialization in control while applying the
+ * JSON-B-specific scalar rules that differ from the generic decoder contract,
+ * such as binary data strategies and untyped {@link BigDecimal} numbers.
+ *
+ * @param delegate The underlying Serde decoder
+ * @param binaryDataStrategy The configured JSON-B binary data strategy
+ */
+record JsonbDecoder(Decoder delegate, String binaryDataStrategy) implements Decoder {
```
dstepanov, line 46: Internal?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbDeserializationContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * JSON-B deserializer callback context backed by the cloned Micronaut runtime mapper.
+ * <p>
+ * Configured JSON-B deserializers can recursively deserialize nested values through this context
+ * while retaining the same JSON-B runtime-introspection and fallback configuration.
+ *
+ * @param codec The bounded runtime codec configured with JSON-B introspection support
+ */
+record JsonbDeserializationContext(JsonbFallbackCodec codec) implements DeserializationContext {
```
dstepanov, line 36: Internal

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbEncoder.java
```diff
@@ -0,0 +1,137 @@
...
+ * JSON-B encoder wrapper around a Serde encoder.
+ * <p>
+ * The wrapper keeps normal Serde serialization in control while applying the
+ * JSON-B-specific scalar rules that differ from the generic encoder contract,
+ * such as binary data strategies and character encoding.
+ *
+ * @param delegate The underlying Serde encoder
+ * @param binaryDataStrategy The configured JSON-B binary data strategy
+ */
+record JsonbEncoder(Encoder delegate, String binaryDataStrategy) implements Encoder {
```
dstepanov, line 39: Internal

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param serializationConfiguration The current serialization configuration
+     * @param deserializationConfiguration The current deserialization configuration
+     * @param jsonbConfiguration The JSON-B integration configuration
+     * @return The JSON-B instance
+     */
+    @SuppressWarnings("unused")
+    @Singleton
+    Jsonb jsonb(JsonbConfig config,
+                BeanContext beanContext,
+                ObjectMapper objectMapper,
```
dstepanov, line 89: We might want to create a custom-static ObjectMapper so it would be possible to use the default ObjectMapper and an enhanced ObjectMapper with JSONB serdes.

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/MicronautJsonbProvider.java
```diff
@@ -0,0 +1,914 @@
...
+import java.util.Optional;
+import java.util.TimeZone;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Micronaut Serialization backed JSON-B provider using generated serializers and deserializers.
+ *
+ * @since 3.1.0
+ */
+public class MicronautJsonbProvider extends JsonbProvider {
```
dstepanov, line 76: Refactor out inner classes

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/MicronautJsonbReflectionProvider.java
```diff
@@ -0,0 +1,815 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+import tools.jackson.core.ObjectReadContext;
+
+/**
+ * Micronaut Serialization backed JSON-B provider with reflection fallback behavior for JSON-B compatibility.
+ *
+ * @since 3.1.0
+ */
+public final class MicronautJsonbReflectionProvider extends MicronautJsonbProvider {
```
dstepanov, line 66: Consider moving out inner classes.
Should it be internal?

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/MicronautJsonbProvider.java
dstepanov: Should it be internal?

dstepanov (changes requested):

dstepanov:
You don't need `SerdeRegistrar` if the serde is loaded from the bean context.

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbTypeDeserializerBridge.java
```diff
@@ -0,0 +1,89 @@
...
+        );
+        JsonbFallbackCodec codec = new JsonbFallbackCodec(
+            mapper,
+            context.getSerdeConfiguration().orElse(null),
+            BinaryDataStrategy.BYTE
+        );
+        return new Deserializer<>() {
+            @Override
+            public @Nullable Object deserializeNullable(Decoder decoder, DecoderContext context, Argument<? super Object> type) throws IOException {
+                if (decoder.decodeNull()) {
```
yawkat: this is default

.github/workflows/json-tck.yml
graemerocher: we need this because we need to publish TCK passing evidence otherwise we can't claim spec compliance.

This will produce a report similar to https://eclipse-ee4j.github.io/odi/tck-results/latest/jdk-25/

it is in a separate workflow precisely because I don't want it to be overwritten by file sync

serde-api/src/main/java/io/micronaut/serde/config/DeserializationConfiguration.java
graemerocher: this is the default in JSON-B yes and that is why we need it. It is a spec compliance requirement

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbFactory.java
graemerocher: There is already one available through `Jsonb jsonb = JsonbBuilder.create();`

graemerocher:
it isn't always loaded through the bean context

.github/workflows/json-tck.yml
dstepanov: It cannot be moved to the Gradle and published on a non-snapshot release?

dstepanov:
> it isn't always loaded through the bean context

Where? I didn't find any cases like that

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
graemerocher: I don't want to couple this PR to a new version of core, it is currently compatible with the release version, we can address that later if necessary with a follow up PR

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
graemerocher: it already exists in InstantiationUtils

serde-jsonb/src/main/java/io/micronaut/serde/jsonb/JsonbBridgeSupport.java
graemerocher: No because that requires dynamic class loading

graemerocher:
> Move all the new JSONB serializers / deserializers / serdes into a new package and have the package condition. Are those discovered by the value or only SERIALIZER_CLASS/DESERIALIZER_CLASS? Maybe we can put `@Bean(exposes=type)`.

I don't understand this comment

serde-jackson/src/main/java/io/micronaut/serde/jackson/JacksonDecoder.java
graemerocher: how would I pick the decoder? Don't think we have an API for that

dstepanov:
You have introduced multiple bean context serdes for common objects like datatime and even a raw object, if you introduce this dependency to an existing project those new serde will override the existing ones.
In most of the use-cases you explicitly define them to be used:
```
                metadata.addAnnotation(SerdeConfig.class.getName(), Map.of(
                    SerdeConfig.SERIALIZER_CLASS, new AnnotationClassValue<>(JsonbZoneOffsetSerde.class),
                    SerdeConfig.DESERIALIZER_CLASS, new AnnotationClassValue<>(JsonbZoneOffsetSerde.class)
                ));
```

We can avoid them being contributed to other places by using `@Bean(typed=JsonbZoneOffsetSerde)`.

The idea with package is probably to complicated, first I thought that the new ObjectMapper is created in a static fashion - not the one we get from the bean context.

.github/workflows/json-tck.yml
graemerocher: no because then we would have the problem of the file sync overwriting this custom logic

graemerocher:
I added @Bean(exposes=type) where I could, adding to some types (like calendar etc.) broke TCK tests

serde-api/src/main/java/io/micronaut/serde/config/SerdeConfiguration.java
graemerocher: the duration one can support it but a global flag is still necessary

serde-api/src/main/java/io/micronaut/serde/ObjectMapper.java
graemerocher: that sounds like a different PR, all this does is make the default object mapper cloneable without adding new APIs

dstepanov (approved):

PR #1355, opened 2026-06-19
CSV Support

khamzatMurj:
This pr add native CSV format support for micronaut-serialization.
with hand of using `JsonContainer : JsonArray & JsonObject` as It fit like `Line-delimited JSON`.
while considering csv either as 
- Matrix i.e `List<List<String>>`  
- `List<Map<String, String>>` when schema existing(Head of csv & enabled through configuration) ;


yawkat:
I don't think this is workable. There are a thousand CSV variants. Every application that produces CSV uses a slightly different approach for quoting, whitespace, special characters etc. See the [issue list](https://github.com/FasterXML/jackson-dataformats-text/issues?q=is%3Aissue%20state%3Aclosed%20label%3Acsv) in jackson CSV, there's so many details to consider and configure. We don't want to open that can of worms, so imo we need to use a CSV library.

PR #1380, opened 2026-08-08
Add micronaut-serde-cbor module for CBOR (RFC 8949)

BryanSant:
Adds a CBOR runtime alongside the existing Jackson Core, JSON-P and BSON format modules, so the same @Serdeable types and generated serdes can be used over a compact binary encoding.

The module uses Jackson's streaming CBOR factory (jackson-dataformat-cbor) purely as a token parser/generator; object mapping stays on Micronaut's build-time serdes. jackson-databind is excluded, so no reflective databind is pulled onto the classpath and the module is native-image friendly. This follows the same approach that micronaut-serde-json uses today.

serde-cbor:
- CborObjectMapper: ObjectMapper over CBORFactory + JacksonEncoder/Decoder, honouring SerdeConfiguration stream limits, views, tree conversion and updateValue. Ordered after the Jackson JSON mapper so it never takes over as the primary JsonMapper.
- CborMessageHandler: MessageBodyHandler/ResponseBodyWriter for application/cbor, ordered below the JSON handler so wildcard negotiation still prefers JSON.
- CborMediaTypes: application/cbor constant, kept local until Micronaut Core gains MediaType.APPLICATION_CBOR.
- SerdeCborConfiguration: micronaut.serde.cbor.* feature toggles, plus write-binary-as-array defaulting to false so byte[] uses native CBOR byte strings (major type 2) instead of the JSON-legacy numeric arrays.

serde-jackson:
- JacksonDecoder now understands VALUE_EMBEDDED_OBJECT, the token binary formats emit for byte strings, in decodeBinary, decodeNode and decodeArbitrary. Previously a byte[] reaching an untyped, tree or buffered path threw "Unexpected token VALUE_EMBEDDED_OBJECT". Trees use base64, the representation JsonNodeEncoder/JsonNodeDecoder already use for binary. JSON parsers never emit this token, so JSON behaviour is unchanged.

Tests: shared basic-serde TCK routed through real CBOR bytes, round-trip, binary fidelity, tree/untyped binary decoding, nesting limits on read and write, updateValue, golden bytes and streaming interop, and an HTTP controller/client test.

Docs: cborQuick guide page with configuration, HTTP usage and known limitations, plus a runnable doc-examples/example-cbor-java sample covering JVM and native image.

Benchmarks: FormatComparisonBenchmark comparing JSON, BSON binary and CBOR on the same payload, and a CborBenchmark micro suite.

Binary compatibility checks are disabled until the module first ships, as was done for serde-properties.


graemerocher:
thanks, could you sign the CLA?

dstepanov:
Thanks @BryanSant

PR #1390, opened 2026-08-26
feat(serde): make float-to-integer coercion configurable

radovanradic:


dstepanov:
Superseded by #1392, which builds on the commits here and generalises the flag into a coercion policy covering every coercion the decoders perform, applied by all of them rather than only `JacksonDecoder`.

The reason for widening the scope: with `accept-float-as-int: false` on this branch, `{"type":"sub","number":42.5}` is rejected but `{"number":42.5,"type":"sub"}` still truncates silently, because the second document buffers `number` until the discriminator resolves the subtype and the buffer is read by `JsonNodeDecoder`. #1392 also fixes a pre-existing case of the same order dependence in the other direction, where `{"number":"42","type":"sub"}` fails with `Not a number` under the default configuration while `{"type":"sub","number":"42"}` returns 42.

`accept-float-as-int` keeps the name and semantics it has here. Closing in favour of #1392.

PR #1398, opened 2026-09-04
feat(serde): make deserialization coercions configurable (#1392)

radovanradic:
This is merge from 3.2.x branch PR #1392


.github/workflows/gradle.yml
```diff
@@ -60,19 +58,19 @@ jobs:
         run: |
           [ -f ./setup.sh ] && ./setup.sh || [ ! -f ./setup.sh ]
 
-      - name: "🚔 Sonatype Scan"
-        if: env.OSS_INDEX_PASSWORD != '' && matrix.java == '17'
-        id: sonatypescan
+      - name: "🚔 Vulnerability Audit"
```
radovanradic, line 61: Switched from sonatype since some transitive dependency was failing and there is no fix available or coming from core

serde-tck-tests/src/main/java/io/micronaut/serde/tck/tests/InetAddressTest.java
```diff
@@ -29,15 +29,9 @@ public class InetAddressTest {
     @Test
     public void inetAddressSerialization(JsonMapper jsonMapper) throws IOException {
         assertEquals(q("127.0.0.1"), jsonMapper.writeValueAsString(InetAddress.getByName("127.0.0.1")));
-        InetAddress input = InetAddress.getByName("google.com");
```
radovanradic, line 32: Micronaut Core 4.10.26 brings Jackson 2.21.5. That Jackson version now validates InetAddress input as an IP literal and rejects hostnames with:

Not a valid IP address string literal

dstepanov:
Closing in favour of a fresh PR to reset the stale CLA check.

PR #1399, opened 2026-09-04
feat(serde): make deserialization coercions configurable (#1392)

dstepanov:
This is merge from 3.2.x branch PR #1392

Replaces #1398 (closed to reset a stale `license/cla` check).


serde-bson/src/main/java/io/micronaut/serde/bson/BsonReaderDecoder.java
```diff
@@ -376,7 +387,8 @@ private byte[] copyValueToDocument() {
     }
 
     private Decoder decoderFromBytes(byte[] documentBytes) throws IOException {
-        BsonReaderDecoder topDecoder = new BsonReaderDecoder(new BsonBinaryReader(ByteBuffer.wrap(documentBytes)), ourLimits());
+        BsonReaderDecoder topDecoder = new BsonReaderDecoder(new BsonBinaryReader(ByteBuffer.wrap(documentBytes)),
+            ourLimits(), getCoercionPolicy());
```
radovanradic, line 391: @dstepanov There is Sonar blocker warning for this. Should it be addressed?
I think it was reported also in 3.2.x but merged like that.

dstepanov: probably

PR #1403, opened 2026-09-07
Experimental Protocol Buffers backend

dstepanov:
Adds `serde-protobuf`, an experimental format module that reads and writes the Protocol Buffers binary wire format for ordinary `@Serdeable` types — no `.proto` files, no generated message classes, no `protoc` in the build.

It targets a different audience from micronaut-grpc, which already covers schema-first development: people who want protobuf on the wire for plain beans over HTTP, Kafka or Pulsar without adopting a `.proto` toolchain.

> **Base branch:** this targets `3.2.x`, not `3.1.x`. The `@since` tags say 3.2, and the branch is rebased onto `3.2.x`.

## What's here

**The module.** An `ObjectMapper`, an encoder, a decoder, a `KeysProvider`, and the wire primitives, modelled on `serde-bson`. `@ProtoField` assigns a field number, `@ProtoType` picks between the representations protobuf offers for a Java type (`int32`, `sint32`, `fixed32`…). Field numbers are validated when a schema is first resolved, so duplicates, out-of-range values and the reserved 19000-19999 range fail with a message naming the property.

**Numbering.** `@ProtoField(3)` and `@ProtoField(position = 3)` are the same thing — `value` is an `@AliasFor` of `position`. Leave the number out and the property takes it from its place among the message's properties; a property with no annotation at all is numbered the same way, so a plain `@Serdeable` record needs nothing added. Explicit and derived numbers can be mixed, and a collision between them is rejected naming both properties.

**Docs and examples.** A guide page whose code comes from a runnable `doc-examples/example-protobuf-java` project, so a change that breaks the documented behaviour breaks the build.

## Two things about the wire format shaped the design

Tags are written **lazily**: serde encodes a key before it knows the value is null, but protobuf represents an absent field by writing nothing, so the tag is only emitted once a value actually arrives. Nested structures **buffer**, because a nested message is length-prefixed and its length is not known until it is finished; those buffers are pooled on the root encoder and reused.

Field numbers are resolved from the type at `encodeObject`/`decodeObject` rather than from the key name, so **no `serde-api` change was needed**. Per-property tags and encoding kinds are computed once when the schema is built, and the name-to-number and number-to-key-index tables once per key set, keeping the per-value path to an integer switch and a write.

## Verification

116 module tests and 8 example tests. Output is checked **byte-for-byte against protobuf-java** across nested messages, packed and unpacked repeated fields, zig-zag and fixed-width scalars and byte strings, through both the generated and the runtime serdes.

Coverage spans every Java scalar a bean might hold — primitives and their boxes, `String`, enums, `BigInteger`/`BigDecimal`, ten date and time types, `UUID`, `URI`, `Locale`, `Charset`, all four `Optional` flavours, extremes, `NaN` and infinity — and `List`, `Set`, `Collection`, `Iterable` and arrays, each checked against the same reference message.

The binary edge cases from the upstream [conformance suite](https://github.com/protocolbuffers/protobuf/tree/main/conformance) are ported as data rather than by running its native test runner: illegal tags, premature end of input, truncated values and packed runs, plus the unusual-but-legal cases — field reordering, last-wins scalars, message merging, repeated fields split across packed and unpacked runs. Porting them caught three real defects in tag handling (field number zero, numbers that don't fit in 32 bits, and overlong tag varints were all being skipped as though merely unknown).

## Not supported

Protobuf payloads are not self-describing — wire type 2 is ambiguous between a string, a byte array, a nested message and a packed repeated field — so schema-free reads cannot work. Reading into `Object` or `JsonNode`, `@JsonAnySetter`, and reactive parsing all fail with an explicit message rather than a wrong answer. Maps, polymorphic types and `Optional` properties are not implemented yet.

Two behavioural differences are worth knowing: proto3 omits fields holding their type's default value, so a reader cannot tell an absent field from an empty string or a zero (use `@Nullable` or `Optional` for explicit presence); and unknown fields are skipped rather than retained, so they are not re-emitted on a read-then-write.

## For reviewers

- **Performance is reasoned about, not measured.** Allocations and lookups were removed from paths that provably run per value, but there is no benchmark. The `benchmarks` module carries a publishing contract feeding the user guide's headline numbers, which did not seem the place for an experimental module. Treat the nested-buffering strategy as unsettled until it is measured.
- **Two pre-existing serde defects surfaced** while sweeping bean shapes, both format-independent (they reproduce through `writeValueToTree`, never reaching an encoder). They are recorded as tests asserting the current failure, so they will turn red when fixed:
  - a `Set` property cannot be deserialized — `ArrayList cannot be cast to Set`, every element type
  - a `byte[][]` property cannot be serialized at all
- **Numbering by declaration order is convenient and fragile.** Inserting or reordering a property renumbers everything after it. The annotation javadoc and the guide both say so, but it is worth a second opinion on whether it should be opt-in rather than the default when no number is given.
- `KeysProvider.createWithMetadata` landed on `3.2.x` and is not used here; the backend gets field numbers from the type instead. Worth deciding whether it should.


serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,558 @@
...
+    private final Class<?> messageType;
+    private final Map<String, ProtoProperty> byName;
+    private final int[] numbers;
+    private final ProtoProperty[] slots;
+    private final Map<Keys, @Nullable ProtoProperty[]> propertiesByKeys = new ConcurrentHashMap<>();
+    private final Map<Keys, int[]> keyIndexesBySlot = new ConcurrentHashMap<>();
+
+    private ProtoSchema(Class<?> messageType, List<ProtoProperty> properties) {
+        this.messageType = messageType;
+        this.byName = HashMap.newHashMap(properties.size());
```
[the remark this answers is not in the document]
dstepanov: Good catch — `HashMap.newHashMap` is JDK 19+, and this module has to build on the same baseline as the rest of the repo.

Replaced with `io.micronaut.core.util.CollectionUtils.newHashMap`, which does the same pre-sizing without raising the baseline. Fixed in b661d2082.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,558 @@
...
+    /**
+     * Give every property a field number.
+     *
+     * <p>Explicit numbers are claimed first, so that a number derived from position can be reported
+     * against the property that already holds it rather than the other way round. Everything left
+     * takes its number from where it sits among the message's properties, counting from one.</p>
+     */
+    private static int[] assignNumbers(Class<?> messageType, List<PropertyDraft> drafts) throws SerdeException {
+        int[] numbers = new int[drafts.size()];
+        Map<Integer, String> owners = HashMap.newHashMap(drafts.size());
```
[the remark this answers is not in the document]
dstepanov: Same as the other `newHashMap` thread — swapped for `io.micronaut.core.util.CollectionUtils.newHashMap` in b661d2082.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,558 @@
...
+            try {
+                return resolve(type);
+            } catch (SerdeException e) {
+                return e;
+            }
+        });
+        if (cached instanceof ProtoSchema schema) {
+            return schema;
+        }
+        throw new SerdeException(String.valueOf(((SerdeException) cached).getMessage()));
```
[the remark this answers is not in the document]
dstepanov: Agreed. The failure is now chained rather than rethrown bare:

```java
throw new SerdeException(String.valueOf(failure.getMessage()), failure);
```

It is still a new exception rather than a rethrow, deliberately: schema resolution is cached per type, so the stored failure carries the stack of whichever call resolved that type first. Rethrowing it would hand every later caller a stack from an unrelated code path. Chaining gives the caller their own stack with the original detail underneath, and there is a comment at the site saying so. Fixed in b661d2082.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufDecoder.java
```diff
@@ -0,0 +1,558 @@
...
+            throw new SerdeException("Expected a numeric protobuf value with wire type " + expectedWireType
+                + " but the field " + describeCurrent() + " has wire type " + actualWireType);
+        }
+        long value = switch (expectedWireType) {
+            case ProtoWire.VARINT -> {
+                long raw = input.readVarint64();
+                yield property != null && property.intKind() == ProtoProperty.KIND_ZIGZAG
+                    ? ProtoWire.zigZagDecode64(raw)
+                    : raw;
+            }
```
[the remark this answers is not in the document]
dstepanov: This was a real decoding bug, not a style point — thank you.

`sint32` is zig-zagged over 32 bits, so decoding all 64 and truncating afterwards gives a different number. With a varint carrying bits above 32 (legal, though no canonical encoder emits it) the old code returned `-2147483647` where protobuf-java gives `1`.

Narrowing now happens before the decode, chosen from the declared type:

```java
yield property.zigZag32()
    ? ProtoWire.zigZagDecode32((int) raw)
    : ProtoWire.zigZagDecode64(raw);
```

Pinned by `BinaryWireFormatConformanceTest.narrowsAnOversizedZigZagVarintBeforeDecodingIt`, alongside an extremes test covering `Integer.MIN_VALUE`/`MAX_VALUE`. Fixed in b661d2082.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+                require(4);
+                position += 4;
+            }
+            case ProtoWire.LENGTH_DELIMITED -> {
+                // read the length first: a compound assignment would capture the position from
+                // before the length varint was consumed
+                int length = readLength(limit);
+                position += length;
+            }
+            case ProtoWire.START_GROUP -> skipGroup(fieldNumber, limit);
```
graemerocher: **High — unbounded recursion on nested groups gives a `StackOverflowError` from a small payload.**

`skip()` → `skipGroup()` → `skip()` recurses once per nested start-group tag with no depth counter. `LimitingStream`'s nesting limit does not apply here — it only guards `decodeObject`. A start-group tag is a single byte (`0x0B`), so recursion depth is roughly the payload length.

Running a payload of N repeated `0x0B` bytes through `ProtoInput.readTag()` + `skip()`:

```
groups n=100     -> EOFException: Truncated protobuf group 1
groups n=10000   -> java.lang.StackOverflowError
groups n=100000  -> java.lang.StackOverflowError
```

This is reachable from `ProtobufMapper.readValue`: `decodeObject` → `ProtoMessageNormalizer.normalize` → `accumulateField` → `input.skip(fieldNumber, wireType, limit)`.

`BinaryWireFormatConformanceTest.rejectsGroupWireTypes` already demonstrates that a start-group tag reaches this code; it passes only because its payloads are short enough to hit EOF before the stack runs out. It also asserts `Exception`, which would not catch the `Error`.

Since groups are proto2-only and unsupported here anyway, the cleanest fix is to reject `START_GROUP` outright alongside `END_GROUP` and delete `skipGroup` — that removes the recursion rather than bounding it:

```java
case ProtoWire.START_GROUP -> throw new IOException("Malformed protobuf payload: group encoding is not supported");
```

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufDecoder.java
```diff
@@ -0,0 +1,564 @@
...
+    @Override
+    public double decodeDouble() throws IOException {
+        expectWireType(ProtoWire.FIXED64, "a double");
+        double value = Double.longBitsToDouble(input.readFixed64());
+        valueConsumed();
+        return value;
+    }
+
+    @Override
+    public BigInteger decodeBigInteger() throws IOException {
```
graemerocher, line 316: **High — `BigInteger`/`BigDecimal` are built from an unbounded, attacker-supplied digit string.**

`new BigInteger(decodeString())` has no length cap, and parsing a decimal string is superlinear. Measured on one core (JDK 21):

| digits | `new BigInteger(...)` |
|---|---|
| 100 000 | 174 ms |
| 500 000 | 3.9 s |
| 1 000 000 | 15.9 s |
| 4 000 000 | 272 s |

So a single ~1 MB payload pins a request thread for ~16 s, and a ~4 MB one for four and a half minutes. Any bean with a `BigInteger` or `BigDecimal` property is exposed; `decodeBigDecimal` below has the same shape.

Jackson caps this by default (`StreamReadConstraints`, 1000 digits) and there is no equivalent here. Suggest checking the string length before constructing the number, ideally against a value carried on `SerdeConfiguration` rather than a hard-coded constant.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufDecoder.java
```diff
@@ -0,0 +1,564 @@
...
+                    decoderContext, childLimits());
+            }
+            case MESSAGE, REPEATED -> {
+                expectWireType(ProtoWire.LENGTH_DELIMITED, "a nested message");
+                int length = input.readLength(limit);
+                int start = input.position();
+                int end = start + length;
+                input.position(end);
+                valueConsumed();
+                byte[] normalized = ProtoMessageNormalizer.normalize(input.buffer(), start, end, childSchema);
```
graemerocher, line 143: **High — every nested message is copied twice, and every ancestor copy stays live, so nesting multiplies heap by the depth limit.**

Each `decodeObject` normalizes the nested message into a **fresh `byte[]`**: once into `FieldAccumulator.combined` (`ProtoMessageNormalizer.accumulateField` copies the whole nested payload with `writeRaw`), then again into the `emit` output. The returned child `ProtobufDecoder` holds that copy, and the parent decoder is still on the stack waiting for `finishStructure`, so **all copies along the current path are strongly reachable at once**.

With the default `LimitingStream.DEFAULT_MAXIMUM_DEPTH` of 1024, and each nesting level costing only a tag byte plus a length varint, a message can stay nearly as large as its parent all the way down. A ~1 MB payload can therefore retain on the order of 1 GB of live heap, plus roughly twice that in churn — around 1000× memory amplification on a request that is otherwise unremarkable. protobuf-java parses in place and limits recursion to 100.

The PR description already flags the nested-buffering strategy as unmeasured; this is the specific reason it is worth measuring before the module stabilises. Options, roughly in order of preference:

1. Parse nested messages in place against a `(start, limit)` window instead of materialising a copy, doing the duplicate/reorder folding lazily.
2. Failing that, track a total normalized-bytes budget across the whole payload rather than per message.
3. As a stopgap, give protobuf a much lower default nesting limit (protobuf-java uses 100), since the cost per level here is far higher than for JSON.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+     * @param limit The position the value must not extend past
+     * @return The length
+     * @throws IOException If the length is negative or reaches past the limit
+     */
+    public int readLength(int limit) throws IOException {
+        int length = readVarint32();
+        if (length < 0) {
+            throw new IOException("Malformed protobuf payload: negative length " + length);
+        }
+        if (position + length > limit) {
```
graemerocher: **Medium — this bounds check is defeated by integer overflow.**

`length` is only checked for being negative, so it can be up to `Integer.MAX_VALUE`; `position + length` then wraps to a negative number and the comparison passes. Confirmed against a **6-byte** payload whose length prefix is `0x7FFFFFFF` (`0A FF FF FF FF 07`):

```
readLength ACCEPTED length=2147483647 on a 6-byte payload
readBytes  -> java.lang.OutOfMemoryError: Requested array size exceeds VM limit
readString -> java.lang.StringIndexOutOfBoundsException: Range [6, 6 + 2147483647) out of bounds for length 6
skip       -> position now -2147483643
             -> java.lang.ArrayIndexOutOfBoundsException: Index -2147483643 out of bounds for length 6
```

Today the `normalize` pass happens to reach the `ArrayIndexOutOfBoundsException` before `readBytes` gets a chance to allocate, so the practical impact is an **unchecked exception escaping a parser that documents `IOException`** — an HTTP 500 rather than a 400, and `MalformedPayloadTest`'s "negative length" guarantee quietly not covering this case. The 2 GB allocation is one refactor away from being reachable.

Subtracting instead of adding fixes it without changing behaviour for valid input:

```java
if (length > limit - position) {
```

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+
+    private byte readByte() throws IOException {
+        if (position >= buffer.length) {
+            throw new EOFException("Truncated protobuf payload");
+        }
+        return buffer[position++];
+    }
+
+    private void require(int bytes) throws IOException {
+        if (position + bytes > buffer.length) {
```
graemerocher: Same overflow as `readLength` above — `position + bytes` wraps negative for a large `bytes`, so the check passes and `readBytes` goes on to `new byte[length]`. Suggest:

```java
if (bytes > buffer.length - position) {
```

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+     * Skip a complete field value, including a deprecated group.
+     *
+     * @param fieldNumber The field number
+     * @param wireType    The wire type
+     * @param limit       The position the value must not extend past
+     * @throws IOException If the payload is truncated or the group is malformed
+     */
+    public void skip(int fieldNumber, int wireType, int limit) throws IOException {
+        switch (wireType) {
+            case ProtoWire.VARINT -> readVarint64();
```
graemerocher: **Medium — the enclosing-structure limit is enforced only for length-delimited values, not for varints or fixed-width values.**

`limit` is honoured in the `LENGTH_DELIMITED` branch, but `readVarint64()` and `require()` bound against `buffer.length` and never see `limit`. So a varint or fixed value inside a nested message can consume bytes that belong to its parent. Confirmed with a nested message that declares length 1 and contains only a `fixed64` tag:

```
skip(FIXED64) ACCEPTED past the nested limit; position now 9 (limit was 1)
skip(VARINT)  ACCEPTED past the nested limit; position now 4 (limit was 1)
readFixed64 across the boundary = 0x8877665544332211
```

Because `normalize` runs on its own `ProtoInput`, the parent's cursor is unaffected and the parent then re-reads those same 8 bytes as its own fields — the same bytes decoded twice with two different meanings. protobuf-java rejects this as truncated input, so it is a parser differential: a payload that a conformant implementation refuses is accepted here, with a different interpretation. That matters wherever something else on the path (a gateway, a validator, a signature check) parses the same bytes.

It also contradicts the invariant the module advertises and that `MalformedPayloadTest.rejectsANestedValueThatReachesPastItsParent` is written to assert — the guarantee currently only holds for one of the three wire shapes.

Fix by threading `limit` through the value reads on this path (a `readVarint64(limit)` / `require(bytes, limit)` overload, or capping the effective end to `min(limit, buffer.length)` while inside a bounded structure).

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+     * Read a length-prefixed UTF-8 string.
+     *
+     * @param limit The position the value must not extend past
+     * @return The string
+     * @throws IOException If the payload is truncated
+     */
+    public String readString(int limit) throws IOException {
+        int length = readLength(limit);
+        require(length);
+        String result = new String(buffer, position, length, StandardCharsets.UTF_8);
```
graemerocher: **Low — malformed UTF-8 is silently replaced rather than rejected.**

`new String(bytes, UTF_8)` substitutes U+FFFD for invalid sequences. The protobuf spec requires `string` fields to be valid UTF-8, and protobuf-java throws `InvalidProtocolBufferException` here. Another cross-implementation differential: two readers of the same bytes end up with different strings, and the corruption is invisible to the application.

A `CharsetDecoder` configured with `CodingErrorAction.REPORT` would match the reference implementation, at the cost of losing the fast `String` constructor path.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
```diff
@@ -0,0 +1,341 @@
...
+            }
+            case ProtoWire.START_GROUP -> skipGroup(fieldNumber, limit);
+            case ProtoWire.END_GROUP -> throw new IOException("Malformed protobuf payload: unexpected end-group tag");
+            default -> throw new IOException("Malformed protobuf payload: unsupported wire type " + wireType);
+        }
+    }
+
+    private void skipGroup(int fieldNumber, int limit) throws IOException {
+        while (position < limit) {
+            int tag = readVarint32();
```
graemerocher: **Low** — tags inside a group are read with `readVarint32()` rather than `readTag()`, so they bypass the validation added elsewhere in this class: field number zero, tags that do not fit in 32 bits, and overlong tag varints are all accepted here. That is exactly the class of defect the PR description says porting the conformance cases uncovered, so it is worth being consistent.

Moot if `START_GROUP` is rejected outright per the comment on line 296.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtoMessageNormalizer.java
```diff
@@ -0,0 +1,167 @@
...
+ */
+final class ProtoMessageNormalizer {
+
+    private ProtoMessageNormalizer() {
+    }
+
+    static byte[] normalize(byte[] source, int start, int limit, ProtoSchema schema) throws IOException {
+        var input = new ProtoInput(source);
+        input.position(start);
+        var fields = new FieldAccumulator[schema.slotCount()];
```
graemerocher, line 41: **Low (allocation amplification)** — an accumulator is allocated for every slot in the schema on every message, including slots the payload never mentions, and each one carries a `ProtoOutput` with a 64-byte array.

For a wide schema that is roughly `slotCount × ~136` bytes per message. A payload made of many empty submessages costs 2 bytes each on the wire and allocates several kilobytes each — around three orders of magnitude of amplification, all garbage. Allocating the accumulator lazily on first use for a slot would remove it.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoOutput.java
```diff
@@ -0,0 +1,297 @@
...
+        int s = size;
+        while ((value & ~0x7FL) != 0) {
+            b[s++] = (byte) ((((int) value) & 0x7F) | 0x80);
+            value >>>= 7;
+        }
+        b[s++] = (byte) value;
+        size = s;
+    }
+
+    private void ensure(int additional) {
```
graemerocher, line 291: **Low** — `size + additional` overflows for very large buffers, making `required > buffer.length` false so the array is not grown and the following `System.arraycopy` throws `ArrayIndexOutOfBoundsException` instead of `OutOfMemoryError`. Only reachable near 2 GB of output, but the same `additional > buffer.length - size` rearrangement suggested for `ProtoInput` applies.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    public <T> @Nullable T readValue(byte[] byteArray, Argument<T> type) throws IOException {
+        ProtobufDecoder decoder = new ProtobufDecoder(byteArray, limits(), decoderContext);
+        return decoderContext.findDeserializer(type)
+            .createSpecific(decoderContext, type)
+            .deserialize(decoder, decoderContext, type);
+    }
+
+    @Override
+    public <T> @Nullable T readValue(InputStream inputStream, Argument<T> type) throws IOException {
+        return readValue(inputStream.readAllBytes(), type);
```
graemerocher, line 139: **Low / docs** — `readAllBytes()` buffers the whole stream with no cap. Over HTTP the server's max request size bounds it, but the module is also aimed at Kafka and Pulsar, where the caller owns that limit. Worth a line in the guide noting that the decoder requires the complete payload in memory and that callers on non-HTTP transports should bound it themselves.

---

graemerocher (commented):
## Security review of the protobuf backend

Reviewed with untrusted payloads in mind — the attack surface is `ProtobufMapper.readValue` on bytes from the network. Findings are inline; this is the summary.

**Nothing structural to worry about.** `ProtoSchema` is derived from declared Java types and never from the wire, so there is no polymorphic-deserialization or gadget surface: no `Class.forName`, no reflection, no type hints carried in the payload. Refusing schema-free reads is the right call and closes off the usual class of problem here. Field-number validation and the enclosing-structure length check are both good instincts, and the ported conformance cases cover truncation and illegal tags well.

The gaps are all in resource exhaustion and boundary enforcement — the areas a binary parser has to get right and where this suite currently has no coverage.

**High**

1. **Unbounded recursion on nested groups** (`ProtoInput:296`) — `skipGroup` recurses once per start-group byte with no depth counter, and `LimitingStream` does not guard it. 10 000 bytes of `0x0B` is a `StackOverflowError`. Reachable from `readValue` through `normalize`. Rejecting `START_GROUP` outright removes it.
2. **Unbounded `BigInteger`/`BigDecimal` parsing** (`ProtobufDecoder:292`) — 1 MB of digits is ~16 s of CPU, 4 MB is ~272 s. No equivalent of Jackson's `StreamReadConstraints`.
3. **Nesting multiplies heap by the depth limit** (`ProtobufDecoder:120`) — each nested message is copied twice by `normalize` and every ancestor copy stays reachable, so at the default depth of 1024 a ~1 MB payload can retain on the order of 1 GB. This is the concrete reason to measure the nested-buffering strategy the PR description already flags as unsettled.

**Medium**

4. **Integer overflow defeats `readLength` and `require`** (`ProtoInput:243`, `:337`) — a 6-byte payload can declare a length of `Integer.MAX_VALUE` and pass both checks. Today it surfaces as an `ArrayIndexOutOfBoundsException` escaping a parser that documents `IOException` (a 500 rather than a 400); the 2 GB allocation is one refactor from being reachable.
5. **The enclosing-structure limit is enforced only for length-delimited values** (`ProtoInput:281`) — varint and fixed-width reads bound against the whole buffer, so a nested message can consume its parent's bytes and the parent then re-reads them as its own fields. protobuf-java rejects this, so it is a parser differential, and it undercuts the invariant `MalformedPayloadTest` asserts.

**Low** — malformed UTF-8 silently replaced rather than rejected (another differential vs protobuf-java); group tags bypassing `readTag` validation; eager per-slot accumulator allocation in `normalize`; overflow in `ProtoOutput.ensure`; uncapped `readAllBytes` worth a docs note for non-HTTP transports.

**Suggested test coverage.** `BinaryWireFormatConformanceTest` is thorough on malformed *structure* but has nothing on resource exhaustion. Worth adding: deeply nested groups, deeply nested messages, an overflowing length prefix, a varint/fixed value crossing its parent's boundary, and an oversized digit string — each asserting `SerdeException`/`IOException` specifically rather than `Exception`, so that an `Error` cannot slip through the assertion.

Findings 1, 2, 4 and 5 were reproduced by running the payloads against `ProtoInput`; the figures quoted inline are measured, not estimated. Finding 3 is from reading the allocation path.

---

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Fixed in 791aac001 by taking your suggestion — `START_GROUP` and `END_GROUP` both throw now, and `skipGroup` is deleted, so the recursion is gone rather than bounded.

You were right that the existing test passed only by accident. `ProtobufWireShapesTest.skipsAnUnknownGroup` asserted the *opposite* behaviour (that a group is skipped), so it had to be replaced rather than kept — it is now `rejectsAGroupRatherThanSkippingIt`. `UntrustedPayloadTest.aStackOfGroupTagsDoesNotExhaustTheStack` runs your 100 000-byte case and asserts on `Throwable`, not `Exception`, so a returning `StackOverflowError` would fail it.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufDecoder.java
dstepanov: Fixed in 791aac001. The digit string is length-checked before the number is constructed, capped at 1000 characters to match Jackson's `StreamReadConstraints` default. `BigDecimal` goes through the same path.

I put the constant in the module rather than on `SerdeConfiguration` as you suggested: `SerdeConfiguration` currently carries only `maximumNestingDepth`, so a length cap there is a new public API across every format, which felt like it wanted its own decision rather than being smuggled in on this PR. Happy to move it if you would rather have it configurable now.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufDecoder.java
dstepanov: Partly fixed in 791aac001, and I want to be clear about which part.

There is now a normalization budget shared across a whole payload — a running total of the bytes spent on normalized copies, capped at `max(1 MiB, 16 × payload length)`. That bounds the amplification regardless of depth, which your option 2 describes. A payload nested so each level is nearly as large as its parent is refused; ordinary nesting is nowhere near the cap. Both cases are tested.

What it does **not** do is remove the copies. Your option 1 — parsing in place against a `(start, limit)` window and folding duplicates lazily — is the actual fix, and it is a real change to how the normalizer and decoder interact rather than something to slip into a hardening commit. The budget stops the bleeding; the redesign is still owed.

I did not take option 3. Lowering the nesting limit would override a value the user configured, and with the budget in place it no longer buys anything.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Fixed in 791aac001 with your `length > limit - position` rearrangement, which is exact for every input rather than merely harder to overflow.

Tested with your exact payload (`0A FF FF FF FF 07`) in `UntrustedPayloadTest.aLengthPrefixNearIntegerMaxDoesNotOverflowTheBoundsCheck`, asserting it fails as a malformed payload rather than an `ArrayIndexOutOfBoundsException` escaping a method that declares `IOException`.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Fixed in 791aac001 — `require` now compares `bytes > limit - position`, and it also takes the enclosing limit rather than always bounding against the buffer, which was the separate problem on the thread below.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Fixed in 791aac001. `readVarint64`, `readFixed32` and `readFixed64` all take a limit now, and `skip` passes the enclosing structure's limit to every branch rather than only the length-delimited one.

Your point about it being a parser differential was the convincing part — thank you for framing it that way. The Codex CLI, run over the same diff as a second opinion, then found the decoding-side twin: a truncated packed run could absorb the field following it, because `readIntegral` and the float/double reads were still unbounded. Both are fixed and tested (`aValueInsideANestedMessageCannotReadPastIt`, `aTruncatedPackedRunCannotConsumeTheFieldAfterIt`, `aPackedRunOfFixedWidthValuesIsBoundedToo`).

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Fixed in 791aac001 using a `CharsetDecoder` with `CodingErrorAction.REPORT`.

On the cost you flagged: ASCII is always valid UTF-8, so a scan for a high bit keeps the fast `String` constructor for the common case and only a payload with non-ASCII pays for the checking decoder. `validNonAsciiStringsStillDecode` covers the accented, CJK and emoji cases so the strict path is not just rejecting everything.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoInput.java
dstepanov: Moot as you predicted — `skipGroup` is gone entirely, so there is no longer a second place that reads tags.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtoMessageNormalizer.java
dstepanov: Fixed in 791aac001 — accumulators are created on first use for a slot, so a message pays only for the fields its payload actually mentions.

Worth recording that I got this wrong on the first attempt: I removed the eager initialisation loop without adding the lazy creation, which NPEs on every non-empty message. The Codex run caught it before I did.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/wire/ProtoOutput.java
dstepanov: Fixed in 791aac001 with the same `additional > buffer.length - size` rearrangement. Agreed it needs ~2 GB of output to reach; it was a one-line change and leaving one of the pair unfixed would just invite the question again later.

serde-protobuf/src/main/java/io/micronaut/serde/protobuf/ProtobufMapper.java
dstepanov: Documented in 791aac001. The guide now has a "Reading untrusted payloads" section that says the decoder needs the whole payload in memory, that HTTP bounds this through the server's max request size, and that callers on Kafka or Pulsar own that limit themselves. It also collects the other guarantees from this review in one place.

graemerocher (approved):

PR #1409, opened 2026-09-08
Add YAML serialization support via snakeyaml-engine

graemerocher:
## Summary
This PR adds comprehensive YAML serialization and deserialization support to Micronaut Serialization using the event-based parser and emitter from snakeyaml-engine. The implementation provides a YAML-backed ObjectMapper that integrates seamlessly with Micronaut's build-time serialization framework.

## Key Changes

- **YamlObjectMapper**: New ObjectMapper implementation for reading and writing YAML documents, exposed as a named bean with the `yaml` qualifier
- **YamlDecoder**: Event-based YAML decoder that handles root mappings, sequences, and scalars with support for YAML 1.2 core schema type resolution
- **YamlEncoder**: Event-based YAML encoder with configurable output formatting (flow style, indentation, quotes, literal blocks, etc.)
- **YAMLAnchorReplayingParser**: Resolves YAML anchors, aliases, and merge keys on top of the low-level event stream, with proper precedence handling
- **SerdeYamlConfiguration**: Configuration class for YAML-specific read/write features including write style, explicit markers, indentation, and quote handling
- **YamlStringQuotingChecker**: Helper to determine when string values and keys need quoting based on YAML reserved keywords
- **YamlMessageHandler**: HTTP body handler for automatic YAML serialization in HTTP requests/responses
- **Comprehensive test suites**: 
  - AbstractYamlDeserializationSpec and AbstractYamlSerializationSpec for TCK tests
  - YamlDecoderSpec, YamlEncoderSpec, YamlObjectMapperSpec for unit tests
  - YamlHttpBodyHandlerSpec for HTTP integration
  - Jackson Databind compatibility tests via AbstractJacksonDatabindYamlSpec
- **Documentation**: Quick start guide for YAML usage with examples
- **Example code**: YamlQuickStartTest and YamlReportService demonstrating YAML mapper usage

## Notable Implementation Details

- Uses snakeyaml-engine's event-based API only as a token source/sink; actual object mapping is performed by Micronaut Serialization's build-time serializers/deserializers
- Supports complex YAML features: anchors, aliases, merge keys, multiple scalar styles, and YAML 1.2 core schema type inference
- Handles edge cases: empty strings vs null, legacy boolean formats, big number precision, unknown nested collections
- Configurable output formatting with sensible defaults (no explicit document markers, minimized quotes, 2-space indentation)
- Proper error handling with clear messages for invalid YAML and unsupported features (e.g., complex mapping keys)

https://claude.ai/code/session_01XoSKiuUVLHpxX3HQ6MNCwZ


graemerocher:
## Security review of the YAML module

Reviewed at `cb6621e`, focusing on the untrusted-input path: `YamlDecoder`, `YAMLAnchorReplayingParser`, `YamlReadSettings` and the HTTP body handler. Claims below were checked against the running code, not just read.

### What holds up well

**No deserialization gadget surface.** This is the property that matters most for a YAML mapper, and the design gets it right. The module only ever touches `org.snakeyaml.engine.v2.api.lowlevel.Parse` — the event stream. There is no `Load`, no `Compose`, no `Constructor` anywhere, so YAML tags never select a Java type; types come exclusively from the build-time serdes. Confirmed by feeding the usual suspects through the mapper:

| Input | Result |
| --- | --- |
| `v: !!javax.script.ScriptEngineManager [[]]` | read as `ArrayList`, nothing instantiated |
| `v: !!java.net.URLClassLoader [[]]` | read as `ArrayList`, nothing instantiated |
| `v: !<tag:yaml.org,2002:int> 7` | `Integer` 7 |

That structurally rules out the SnakeYAML gadget class of issue (CVE-2022-1471 and relatives). Worth stating explicitly in the guide, since it is the first question a security reviewer asks of any YAML dependency.

Also good: `micronaut.serde.maximum-nesting-depth` is enforced (verified — a low limit raises `SerdeException`); recursive anchors are rejected rather than expanded; multi-document input is rejected; the parser and the skip path are iterative, so no stack-overflow vector; reference-counted buffers are released when decoding fails; and the encode-side `CodecException` names only the object's type, not its value.

### Finding 1 — alias expansion has no cumulative budget (denial of service)

`MAX_EVENTS` (9999) bounds the replay queue *at any instant* and each anchor's recording, but nothing bounds the *total* number of replayed events across a document. The queue drains as events are consumed, so N aliases each expanding to ~9999 events all pass the check individually.

Measured on this branch, reading into a type that ignores the amplified key:

| Document | Events processed | Time |
| --- | --- | --- |
| 1.4 KB | 1.8 M | 168 ms |
| 3.8 KB | 7.3 M | 481 ms |
| 13.4 KB | 29.1 M | 1129 ms |

That is roughly 2,200 events per input byte, scaling linearly. At the default 3 MiB `code-point-limit` that extrapolates to ~7 × 10⁹ events, i.e. minutes of single-threaded CPU from one request. Shape of the input:

```yaml
l0: &l0 [x, x, ...]        # ~100 scalars
l1: &l1 [*l0, *l0, ...]    # ~90 aliases -> ~9000 events, just under the cap
data: [*l1, *l1, *l1, ...] # fan-out bounded only by document size
```

Suggested hardening: a per-document expansion budget — increment a counter in `expandAlias` by `recorded.size()` and fail once the total crosses a threshold (a multiple of the input size, or a configurable `max-alias-expansions`). Jackson's `YAMLAnchorReplayingFactory` carries the same shape of per-alias-only limits, so this is not a regression against the reference implementation, but a cumulative budget is what actually closes it.

### Finding 2 — anchor retention multiplies two independent caps (memory exhaustion)

`MAX_REFS` caps the number of anchors at 9999 and `MAX_EVENTS` caps each anchor's recording at 9999, but the two multiply: up to ~10⁸ retained event references. Measured: 900 anchors defined by a 16 KB document retained ~126 MB of heap, which extrapolates to roughly 0.7–1.4 GB from a document under 200 KB.

Suggested hardening: cap total recorded events across all anchors rather than per anchor.

### Finding 3 — `code-point-limit` default is generous for untrusted input

3 MiB is the only input-size control inside the mapper, and it is the denominator for both findings above. Worth documenting a lower value for HTTP-facing use, or defaulting the message-body path lower than the programmatic one.

### Finding 4 — media type matching accepts `*/*` (low)

`YamlMediaTypes.isYaml(MediaType.ALL_TYPE)` returns `true`, so `isReadable`/`isWriteable` claim wildcard requests. The CBOR handler does the same via `matchesAllOrWildcardOrExtension`, so this may be intentional, but it is worth confirming that JSON still wins negotiation for `Accept: */*` given `@Order(10)`.

### Finding 5 — decode errors carry document content (informational)

`InvalidFormatException` carries the offending scalar and decoder errors append line and column; `YamlMessageHandler.decorateRead` folds `e.getMessage()` into the `CodecException`. For a request body that is the caller's own data, but the same mapper reads files and config, where a propagated message could disclose content. Worth checking that the application's error handler does not echo it verbatim.

### Scope and caveats

Findings 1 and 2 are availability issues on untrusted input, not integrity or confidentiality ones; neither allows code execution or type confusion. Both predate the current branch in shape — they come with anchor and alias support itself — and neither blocks the feature so long as the limits are documented for untrusted use.

CI has still not run on this PR (all jobs queued), so everything above, including the measurements, comes from local runs on JDK 25.

---

dstepanov (approved):

PR #1412, opened 2026-09-09
Broaden build-time serializer and deserializer generation

graemerocher:
## Summary

`SimpleSerdeShapeAnalyzer` routed most real-world types to the runtime introspection serdes. The two largest causes were heuristics rather than features: any bean with a camel-case property name, or with three or more single-letter fields out of alphabetical order, fell back so that a globally configured naming strategy or alphabetical sort could never disagree with generated code. Beyond that, every Jackson annotation was blacklisted, so `@JsonProperty("x")`, `@JsonIgnore`, `@JsonIgnoreProperties`, `@JsonInclude`, `@JsonPropertyOrder`, `required` and `@JsonAlias` each meant the runtime path; a bean with one getter-only or setter-only property was rejected entirely; and a class without a no-arg constructor was an unsupported shape.

This PR widens generation in focused commits, each with parity specs comparing the generated path against the runtime serdes (and Jackson Databind where semantics agree):

1. **perf**: the naming/ordering heuristics are removed. A runtime property naming strategy or alphabetical sorting is detected once when the specific serde is created (`GeneratedSerdeFallbackUtil`) and routes to the runtime serde; the hot path is unchanged.
2. **feat**: renamed properties (`@JsonProperty`, compile-time naming strategies). The source-gen visitor now runs after `SerdeAnnotationVisitor` (visitors execute from highest order to lowest), so generated code observes the property metadata that visitor adds.
3. **feat**: ignored, read-only, write-only, getter-only and setter-only bean properties; type-level `@JsonIgnoreProperties` / `@JsonIncludeProperties`; `ignoreUnknown` as a build-time constant. Bean properties are resolved with the query the introspection visitor uses, because the cached property list predates the visitor's metadata rewrites.
4. **fix**: abstract types and subclasses of a subtyped hierarchy stay on the runtime; `DefaultSerdeRegistry` no longer picks generated serdes for a supertype through covariance.
5. **feat**: `@JsonPropertyOrder`, explicit and alphabetic, in write order matching the runtime.
6. **feat**: `@JsonInclude` on a type, a property or a package, compiled to the check matching the serde that writes the value (`NON_NULL` becomes a null check; a serializer whose inclusions are all declared never consults configuration). Content inclusion still falls back.
7. **feat**: `@JsonProperty(required = true)` and `@JsonAlias`; record deserializers also honor `require-all-creator-parameters`.
8. **feat**: immutable classes bound through their constructor are generated with the record generators.
9. **fix**: parity gaps the wider generation exposed in the module suites: serialize-only imports, argument-level ignore lists and BSON discriminators, type-variable names in argument text, and runtime-identical messages for failing constructors, getters and null constructor parameters.

Still on the runtime path by design: subtyped hierarchies, `@JsonUnwrapped`, any-getter/any-setter, builders, content inclusion, formats and feature overrides, custom serdes, member-level order/ignore lists, `@Bindable` defaults, `@JsonIgnore` on record components.

## Verification

- `:micronaut-serde-jackson:test` (1749 tests), plus the `stax-xml`, `jsonp`, `bson`, `jackson-cbor`, `yaml`, `woodstox-xml` and `oracle-jdbc-json` suites: all green.
- Checkstyle on `serde-api`, `serde-support`, `serde-processor`; `spotlessCheck`.
- Benchmarks: `PropertyAccessShapeBenchmark` and `UserBeanSerdeBenchmark`, generated and runtime stacks, a baseline JMH jar from `3.2.x` versus this branch on the same JVM. All deltas within run-to-run noise (±3–5%), no directional regression. A try/catch initially added around the bean no-arg constructor showed as a small slowdown on generated getter/setter deserialization and was removed; records and constructor-bound classes keep the wrapping.

## Reviewer notes

- The visitor order change (commit 2) is the one cross-cutting behavioral change: generated serdes now see everything `SerdeAnnotationVisitor` adds programmatically, which is what makes naming strategies, ignore lists and package-level inclusion visible to the generator.
- `SOURCEGEN_SHAPE` gains a `CONSTRUCTOR_BEAN` value.
- No user-guide changes: the guide has no section describing generated-serde eligibility, and the benchmark README numbers were not refreshed because there is no change to publish.


serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/records/RecordDeserializerSourceGen.java
```diff
@@ -938,20 +1048,107 @@ private StatementDef deserializeComponentDirect(VariableDef.This aThis,
...
+        if (reference && !usesNonNullScalarDecode) {
+            // The runtime rejects an explicit null for a required parameter, nullable or not
             deserializeAndAssign = StatementDef.multi(
                 deserializeAndAssign,
-                valueVariable.isNull().ifTrue(
-                    nullValueOrDispatchStatement(type, argumentExpression, dispatchResultVariable)
+                valueVariable.isNull().and(requiredComponentCondition(aThis, component)).ifTrue(
+                    requiredConstructorParameterStatement(type, argumentExpression, index)
                 )
             );
```
dstepanov, line 1047: The comment states the opposite of what the runtime does: `DeserBean.deserializeConstructorValue` returns the value *before* it ever tests `explicitlyRequiredForConstructor`:

```java
// DeserBean.java:1794
if (value != null || nullable) {
    return value;
}
if (explicitlyRequiredForConstructor) { throw ... }
```

`nullable` is `argument.isNullable()` (DeserBean.java:1248), so a `@Nullable` component with an explicit null is accepted by the runtime and rejected here.

Repro: `record R(@Nullable String name, int age)` with `require-all-creator-parameters: true` (or `@JsonProperty(required = true)` on the component) and `{"name":null,"age":1}` — the runtime serde constructs `R(null, 1)`, the generated one throws `Required constructor parameter [...] is not present or is null`.

The guard needs `&& !component.nullable()`, matching the `nonNull() && !nullable()` pattern used by the block just below.

---

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/beans/BeanDeserializerSourceGen.java
```diff
@@ -978,7 +1161,12 @@ private StatementDef deserializeAndAssignProperty(VariableDef.This aThis,
             deserializeAndAssign = deserializedValueDef;
             StatementDef assignStatement = assignProperty(beanVariable, property, deserializedValueDef.variable());
             StatementDef propertyAssignment;
-            if (property.nonNull() && !property.nullable()) {
+            if (property.required()) {
+                propertyAssignment = deserializedValueDef.variable().isNull().ifTrue(
+                    requiredPropertyNullStatement(type, argumentExpression),
+                    assignStatement
+                );
+            } else if (property.nonNull() && !property.nullable()) {
```
dstepanov: Same `@Nullable` parity gap as on the record path. `DeserBean.deserializeValue` short-circuits on `nullable` before the required check:

```java
// DeserBean.java:1563
if (value != null || nullable) {
    return value;
}
if (explicitlyRequired) { throw ... }
```

So `@Nullable @JsonProperty(required = true) String name` with `{"name":null}` sets `null` on the runtime serde and throws on the generated one.

Note the `else if` immediately below already gets this right (`property.nonNull() && !property.nullable()`) — the `property.required()` branch just needs the same `&& !property.nullable()`. The equivalent branches at lines 1117, 1221 and 1268 need it too.

---

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/SimpleSerdeShapeAnalyzer.java
```diff
@@ -600,6 +674,17 @@ private boolean isDeserializerSkipped(ClassElement element) {
...
+        Set<String> serializedNames = new HashSet<>();
+        for (PropertyElement property : element.getBeanProperties()) {
+            String serializedName = property.stringValue(SerdeConfig.class, SerdeConfig.PROPERTY).orElse(property.getName());
+            if (!serializedNames.add(serializedName)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
```
dstepanov, line 664: This reads the cached `element.getBeanProperties()`, but `BeanSerdeShapeResolver.introspectedProperties` deliberately re-resolves with `PropertyElementQuery.of(element)` — per the PR description, "because the cached property list predates the visitor's metadata rewrites".

That asymmetry matters most here, since this is the guard the generated `Keys`/switch layout relies on: if the query-resolved list contains a property the cached list does not, a duplicate serialized name goes undetected, the shape is accepted, and `keysCreateExpression` is built from a name list that collapses two properties into one index.

The other eligibility scans in this class have the same staleness exposure (lines 397, 412, 430, 490, 594, 689, 746, 777, 828, 843, 879, 893, 915, 962). Routing them all through `BeanSerdeShapeResolver.introspectedProperties` would keep the analyzer and the resolver looking at the same property set.

---

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/records/RecordDeserializerSourceGen.java
```diff
@@ -907,6 +998,25 @@ private StatementDef deserializeComponentDirect(VariableDef.This aThis,
...
+                            valueVariable.assign(nullableValueDef.variable()),
+                            requiredComponentCondition(aThis, component).ifTrue(requiredConstructorParameterStatement(type, argumentExpression, index))
+                        )
+                    );
+                } else {
+                    keepDefaultOnNullStatement = objectDecoder.invoke(DECODE_NULL_METHOD).isTrue().doIfElse(
+                        requiredComponentCondition(aThis, component).ifTrue(requiredConstructorParameterStatement(type, argumentExpression, index)),
+                        valueVariable.assign(objectDecoder.invoke(scalarDecodeMethod))
+                    );
+                }
```
dstepanov, line 1006: This block unconditionally overwrites the `keepDefaultOnNullStatement` assigned by the structurally identical `if (useNullableScalarDecodeForDefaultPrimitive(...))` / `else` just above (lines 987-1000), making that one dead. The dead block also builds a local via `RecordSerdeSourceGenUtils.localName("decodedValue", index)` with the same generated name as the live one here, so the two would collide if it ever became reachable again.

Looks like the earlier block should have been replaced rather than left in place — worth deleting lines 987-1000.

---

dstepanov (commented):
Reviewed the full non-test diff, cross-checking each generated path against its runtime equivalent in `DeserBean`, `SerBean`, `CustomizedObjectSerializer` and `DefaultSerdeRegistry`.

Two confirmed parity bugs, both the same root cause: the generated required-property null check fires unconditionally, while the runtime returns the null first (`if (value != null || nullable) return value;` precedes the `explicitlyRequired` throw in both `deserializeValue` and `deserializeConstructorValue`). So `@Nullable` + `required`/`require-all-creator-parameters` diverges between the two stacks. Inline on `RecordDeserializerSourceGen` and `BeanDeserializerSourceGen`.

Also flagged: the eligibility scans in `SimpleSerdeShapeAnalyzer` still read the cached `getBeanProperties()` that `BeanSerdeShapeResolver` deliberately bypasses, and a dead `keepDefaultOnNullStatement` block in the record generator.

Checked and found correct: visitor ordering (`+50` vs `SerdeAnnotationVisitor`'s `+100`), key/alias/ignored index layout in both `Keys` construction and switch dispatch, `USE_DEFAULTS` to runtime-config mapping, `NON_EMPTY`/`NON_DEFAULT` build-time inclusion against `Serializer.isEmpty` overrides, property-order parity with `SerBean.applyPropertyOrder`, `ignoreUnknown` resolution, the lazy-copy in `DefaultSerdeRegistry.withoutSpecificSerdesForOtherTypes`, and the new exception message formats against their runtime originals.

---

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/records/RecordDeserializerSourceGen.java
graemerocher: Fixed in 26f839b73: the record path now skips the required-null check for a nullable component (`&& !component.nullable()`), matching `deserializeConstructorValue`. Note the runtime's `SpecificObjectDeserializer` (used when a record declares aliases) still routes an explicit null through `setDefaultConstructorValue` and throws; `GeneratedRequiredPropertyParitySpec` documents that divergence.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/beans/BeanDeserializerSourceGen.java
graemerocher: Fixed in 26f839b73: all four `property.required()` branches in the bean generator now require `!property.nullable()`, so a nullable required property accepts an explicit null like `deserializeValue`. Covered by the new `SourceGenNullableRequiredBean` rows in `GeneratedRequiredPropertyParitySpec`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/SimpleSerdeShapeAnalyzer.java
graemerocher: Fixed in 26f839b73: every property scan in `SimpleSerdeShapeAnalyzer` (and the two in `RecordSerdeShapeResolver`) now goes through `BeanSerdeShapeResolver.introspectedProperties`, memoized per analyzed type, so the eligibility checks and the shape resolvers look at the same property set.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

serde-processor/src/main/java/io/micronaut/serde/processor/sourcegen/records/RecordDeserializerSourceGen.java
graemerocher: Removed in 591b09a86: the dead `keepDefaultOnNullStatement` block is gone, only the live branch remains.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Issue #1254, opened 2025-11-17
Support Token-Oriented Object Notation (TOON)


sdelamo:
https://github.com/toon-format/toon

realbucksavage:
Is it still on the MN roadmap? I could pick this up since we have a usecase where we want to build a service that can spit out toon. There are workarounds by using toon-java but a native MN feature looks nice to have.

graemerocher:
contributions welcome

realbucksavage:
@graemerocher do you think a TCK will be needed for this support?

graemerocher:
well the TCKs are normally needed to test parity with Jackson, but not sure if Jackson supports TOON

realbucksavage:
It doesn't. However, there is https://github.com/toon-format/toon-java which appears to be the Jackson of toon world. So TCK in this case could test the parity whith this.

graemerocher:
ok makes sense

realbucksavage:
Addressed through https://github.com/micronaut-projects/micronaut-serialization/pull/1414.

PR #1414, opened 2026-09-10
Introduce serde capability for TOON

realbucksavage:
Created from #1254


serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonEncoder.java
```diff
@@ -0,0 +1,410 @@
...
+                } else {
+                    consumer.accept(line);
+                }
+            }
+        };
+
+        if (item.isObject()) {
+            writeObjectFields(itemConsumer, item, depth + 1);
+        } else {
+            writeArrayNode(itemConsumer, null, item, depth + 1);
```
graemerocher, line 218: **Blocking.** For a bare array in list-item position the spec (and fixture `arrays-nested.json` / "encodes root-level array mixing primitive, object, and array of objects in list form") wants the nested items one level below the `- ` line, not two:

```
[3]:
  - summary
  - id: 1
    name: Ada
  - [2]:
    - id: 2
    - status: draft
```

This branch emits the inner items at 6 spaces. jtoon rejects that: `Array length mismatch: declared 2, found 0`. The `depth + 1` here is right for the *header* line (which the wrapper consumer then de-indents), but the body is rendered relative to `depth + 1` as well, so it ends up at `depth + 2`. Note the §9.4 example with `- users[2]{id,name}:` (array as the *first field of an object item*) legitimately has rows at `depth + 2` — it's only the keyless `- [N]:` case that is different.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonEncoder.java
```diff
@@ -0,0 +1,410 @@
...
+        int size = array.size();
+        if (size == 0) {
+            consumer.accept(key == null ? prefix + "[]" : prefix + ": []");
+            return;
+        }
+
+        List<JsonNode> elements = CollectionUtils.iterableToList(array.values());
+        if (allPrimitive(elements)) {
+            writeInlineArray(consumer, prefix, elements);
+        } else if (isTabularEligible(elements)) {
```
graemerocher, line 145: **Blocking.** Per §9.3 / §6, tabular form is only valid at document root or in object-field position; an array in list-item position must use list form. `writeListItem` routes here via `writeArrayNode(itemConsumer, null, item, …)` with `key == null`, so `List<List<Bean>>` produces

```
a[1]:
  - [2]{x}:
      1
      2
```

which jtoon rejects (`Keyless array header with field list only valid at document root at line 2`). Spec fixture: `arrays-nested.json` / "uses list form for a tabular-eligible array in list-item position". Simplest fix is to skip the tabular check when `key == null && depth > 0` (root is fine).

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonEncoder.java
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * An array is tabular-eligible when it has at least two elements, every
+     * element is a non-empty object with the same keys in the same order,
+     * and every column is a uniform-primitive column or itself a uniform,
+     * tabular-eligible column of nested objects. A single element is never
+     * eligible, so an ordinary single-field nested object or single-entry
+     * map does not encode as a one-row tabular block.
+     */
+    private boolean isTabularEligible(List<JsonNode> elements) {
+        if (elements.size() < 2) {
```
graemerocher, line 243: §9.3 has no minimum element count for tabular arrays — only §9.5 (keyed objects) requires at least two entries. So `[{id:1,name:"A"}]` should encode as `items[1]{id,name}:\n  1,A` (jtoon does exactly that). Three encode fixtures fail on this (`arrays-tabular.json` "collapses sibling nested field groups…", "quotes subfield names…", `arrays-objects.json` "encodes a keyed-eligible object in a tabular column…").

Since `isKeyedTabularEligible` delegates here, you'll need to keep the `< 2` guard for the keyed path and drop it for arrays — the javadoc rationale about a single-entry map is right for keyed form, just not for arrays.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonEncoder.java
```diff
@@ -0,0 +1,410 @@
...
+        List<String> fieldOrder = List.of();
+        for (JsonNode element : elements) {
+            if (!element.isObject() || element.size() == 0) {
+                return false;
+            }
+
+            List<String> keys = keysOf(element);
+            if (fieldOrder.isEmpty()) {
+                fieldOrder = keys;
+            } else if (!fieldOrder.equals(keys)) {
```
graemerocher, line 256: §9.3: "all objects share the same key *set* (order may vary per object)", header order taken from the first object, and each row's cells emitted in header order. Same for §9.5 keyed objects. Requiring identical order here means `[{a:1,b:2},{b:20,a:10}]` falls back to list form, whereas the canonical output (and what jtoon emits) is

```
items[2]{a,b}:
  1,2
  10,20
```

Fixtures: `arrays-objects.json` "uses field order from first object for tabular headers", `objects-keyed.json` "orders fields by the first entry value's encounter order". `buildRow` already looks cells up by name, so this is just relaxing the equality check to a set comparison (plus the nested-group recursion).

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+            try {
+                int end = ToonEscapes.findClosingQuote(content, 0);
+                String key = ToonEscapes.unquote(content.substring(0, end + 1));
+                return new KeyResult(key, content.substring(end + 1));
+            } catch (SerdeException e) {
+                throw annotateWithLine(e, lineNumber);
+            }
+        }
+
+        if (!ToonEscapes.isKeyStart(content.charAt(0))) {
```
graemerocher, line 421: §7.4 (normative): *"an unquoted key token is the text before the first unquoted colon of a key-value line, with surrounding spaces trimmed; the text before a header's bracket segment; or a field name in a field list. Decoders MUST accept any such token as a literal key, in strict and non-strict mode alike, even when it does not match §7.3's unquoted-key pattern."*

So `foo-bar: 1`, `2key: x`, `foo-bar[2]: 1,2` and `k[2]: 5` inside a keyed block must all be accepted (5 decode fixtures in `objects.json` / `objects-keyed.json` fail here; jtoon accepts all of them). Reading the key as "scan to the first unquoted `:` or `[`" rather than by the §7.3 character class fixes it. The `isKeyStart`/`isKeyPart` grammar is the *encoder's* quoting rule, not a decoder constraint.

This matters more for TOON than most formats — hand-written / LLM-generated input with `Content-Type`-style keys is a primary use case.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+                while (i < s.length() && s.charAt(i) != delimiter && s.charAt(i) != '{' && s.charAt(i) != '}') {
+                    i++;
+                }
+
+                if (i == nameStart) {
+                    throw new SerdeException("Malformed field list (empty field name) at line " + lineNumber + ": " + s);
+                }
+
+                name = s.substring(nameStart, i);
+                if (!ToonEscapes.isValidUnquotedKey(name)) {
```
graemerocher, line 609: Same §7.4 point for field lists: `[1]{2key}:` must be accepted (`arrays-tabular.json` / "accepts a field name outside the encoder unquoted-key pattern"). The scan above already stops at the delimiter / `{` / `}`, so this validation can simply go.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+            cursor++;
+
+            KeyResult keyResult = readKey(entryLine.content(), entryLine.lineNumber());
+            if (keyResult.remainder().isEmpty() || keyResult.remainder().charAt(0) != ':') {
+                throw new SerdeException("Malformed keyed tabular entry at line " + entryLine.lineNumber() + " (expected 'key: values'): " + entryLine.content());
+            }
+
+            List<String> cells;
+            try {
+                cells = ToonEscapes.splitByDelimiter(keyResult.remainder().substring(1), header.delimiter());
```
graemerocher, line 298: An entry row with nothing after the key — `m[1:]{v}:\n  a:` — currently yields `{m: {a: {v: ""}}}` because `splitByDelimiter("")` returns one empty token, which happens to equal `leafCount == 1`. §14.1 wants this to be an error (`validation-errors.json` / "throws on an entry row with no cells after the entry key"). Probably worth rejecting an empty (post-trim) remainder here explicitly.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+        if (!text.isEmpty() && text.charAt(0) == BOM) {
+            text = text.substring(1);
+        }
+
+        List<Line> parsedLines = new ArrayList<>();
+        int lineNumber = 0;
+        for (String rawLine : LINE_SPLIT.split(text, -1)) {
+            lineNumber++;
+            String line = ToonEscapes.stripTrailingWhitespace(rawLine);
+            if (line.isBlank()) {
```
graemerocher, line 91: Strict-mode nit (lower priority): §12/§14.2 say blank lines *inside* an array body (list items, tabular rows, keyed entry rows) MUST error in strict mode; dropping them in the pre-pass makes 9 `blank-lines.json` fixtures pass silently, and jtoon errors on them (`Blank line inside list array at line 3`). Being lenient on input is a defensible choice, but then the class javadoc ("Operates in the specification's strict mode") should say so. If you do want strict, keeping blank lines as `Line` records with a flag and rejecting them inside a declared-count block is probably the least invasive way.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+     * requiring every further sibling to match it exactly.
+     */
+    private void continueObjectFields(int parentIndent, Map<String, JsonNode> values) throws SerdeException {
+        if (cursor >= lines.size() || lines.get(cursor).indent() <= parentIndent) {
+            return;
+        }
+
+        increaseDepth();
+        try {
+            int fieldsIndent = lines.get(cursor).indent();
```
graemerocher, line 167: Related strict-mode nit: because indentation is inferred rather than checked against `indentSize`, `a:\n   b: 1` (3 spaces) and `a:\n    b: 1` (depth jump) are accepted; §12 says both MUST error in strict mode and jtoon rejects them (`Non-multiple indentation: 3 leadingSpaces with indent=2`). 5 `indentation-errors.json` fixtures. Not blocking — but if we keep inference, the javadoc should describe it as more lenient than strict mode rather than claiming strict.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonDocumentParser.java
```diff
@@ -0,0 +1,810 @@
...
+            return JsonNode.createArrayNode(List.of());
+        }
+
+        if (looksLikeScalarAtLine(rest, itemLine.lineNumber())) {
+            return parseScalarToken(rest, itemLine.lineNumber());
+        }
+
+        // A list item that is itself an array will have its depth charged
+        // in parseArrayOrKeyedBody.
+        if (rest.startsWith("[")) {
```
graemerocher, line 355: §6: a keyless header *with* a field list (`- [2]{x}:` or `- [2:]{v}:`) is only valid at the document root, not as a list item; two `validation-errors.json` fixtures expect an error here. Worth rejecting `header.fields() != null` in this position, particularly once the encoder side stops emitting it.

serde-toon/src/main/java/io/micronaut/serde/toon/util/ToonEscapes.java
```diff
@@ -0,0 +1,438 @@
...
+                }
+
+                case 'u' -> {
+                    if (i + 6 > inner.length()) {
+                        throw new SerdeException("Invalid \\u escape sequence: " + quoted);
+                    }
+
+                    String hex = inner.substring(i + 2, i + 6);
+                    try {
+                        sb.append((char) Integer.parseInt(hex, 16));
```
graemerocher, line 199: Two small robustness points on `\u` escapes:
- `Integer.parseInt(hex, 16)` accepts a leading sign, so `\u+041` decodes to `A` and `\u-041` produces garbage; better to require 4 hex digits.
- §7.1: a lone surrogate (`\uD800` not followed by a low surrogate, or vice-versa) MUST be rejected. Fixture `validation-errors.json` / "throws on lone surrogate code point \uD800" currently decodes to a string containing an unpaired surrogate.

serde-toon/src/main/java/io/micronaut/serde/toon/ToonMapper.java
```diff
@@ -0,0 +1,270 @@
...
+ * from that tree. Serializes to a {@link JsonNode} tree and writes it as a
+ * TOON document.</p>
+ *
+ * @see <a href="https://github.com/toon-format/spec">TOON specification</a>
+ * @since 3.2.0
+ */
+@Singleton
+@Secondary
+@Named(ToonMapper.NAME)
+public final class ToonMapper implements ObjectMapper {
```
graemerocher, line 58: `ObjectMapper`'s defaults make a few things silently no-ops here:

- `cloneWithConfiguration(...)` returns `this`, so callers that clone the mapper with a different `SerdeConfiguration` / `SerializationConfiguration` / `DeserializationConfiguration` get the original configuration back with no error.
- `cloneWithViewClass` / `@JsonView` isn't supported (the contexts are created with a `null` view).
- `updateValue*` / `updateValueFromTree` aren't supported.

`YamlObjectMapper` on this branch implements all three (it threads `view` and the cloned registry through a private constructor and uses `JsonViewUtil.extractView`), so it's a good template. I'd consider at least `cloneWithConfiguration` and view support in-scope for a new module.

serde-toon/src/main/java/io/micronaut/serde/toon/ToonMapper.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param type The target type
+     * @param <T>  Type variable of the return type
+     * @return The deserialized value
+     * @throws IOException If tree decoding fails
+     */
+    @Override
+    public <T> @Nullable T readValueFromTree(JsonNode tree, Argument<T> type) throws IOException {
+        try (var decoderContext = registry.newDecoderContext(null)) {
+            Deserializer<? extends T> deserializer = decoderContext.findDeserializer(type).createSpecific(decoderContext, type);
+            return deserializer.deserializeNullable(JsonNodeDecoder.create(tree, limits()), decoderContext, type);
```
graemerocher, line 116: Minor: every other tree-backed mapper passes the configured coercion policy here (`JsonNodeDecoder.create(tree, limits, coercionPolicy)`), resolved the way `YamlObjectMapper` does in its constructor via `CoercionPolicy.fromConfiguration(context.getDeserializationConfiguration())`. Without it `micronaut.serde.deserialization.coercion` has no effect for TOON.

serde-toon/src/main/java/io/micronaut/serde/toon/SerdeToonConfiguration.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * TOON-specific configuration.
+ *
+ * @since 3.2.0
+ */
+@BootstrapContextCompatible
+@Internal
```
graemerocher, line 35: Nit: this is the public, user-facing configuration type (`micronaut.serde.format.toon.*`) and is referenced from the docs, so `@Internal` doesn't seem right — `SerdeYamlConfiguration` isn't marked internal. `@Internal` on the nested `CharSequenceToDelimiterConverter` is fine.

src/main/docs/guide/quickStart/toonQuick.adoc
```diff
@@ -0,0 +1,75 @@
+Micronaut Serialization includes support for reading and writing https://github.com/toon-format/spec[TOON] (Token-Oriented Object Notation) documents with the `micronaut-serde-toon` module.
+
+Add the following artifact to the `dependencies` block:
+
+dependency:micronaut-serde-toon[groupId="io.micronaut.serde",scope="implementation"]
+
+The TOON mapper is exposed as a named api:serde.ObjectMapper[] bean. Inject it with the `toon` qualifier, or by its concrete type api:serde.toon.ToonMapper[], when you want to read or write TOON data:
+
+[source,java]
```
graemerocher, line 9: Nit: elsewhere in this guide (see `yamlQuick.adoc`) examples are `snippet::` includes from `doc-examples/example` so they compile and run in CI (including the native-image matrix). A small `ToonReportService` / `ToonQuickStartTest` pair there would be consistent and gives us a native smoke test for free.

Also worth mentioning in this section what a caller can expect: the mapper is tree-based (whole document read into a `JsonNode` before deserialization), and — until the points on `ToonMapper` are addressed — that views / `cloneWithConfiguration` aren't supported.

serde-toon-tck/src/main/groovy/io/micronaut/serde/toon/tck/AbstractToonJToonParitySpec.groovy
```diff
@@ -0,0 +1,84 @@
...
+package io.micronaut.serde.toon.tck
+
+import dev.toonformat.jtoon.JToon
+import io.micronaut.core.type.Argument
+import spock.lang.Specification
+
+/**
+ * Validates spec compatibility and round-trip parity against the reference {@code dev.toonformat:jtoon} (toon-java) implementation.
+ */
+abstract class AbstractToonJToonParitySpec extends Specification implements ToonSpec {
```
graemerocher, line 25: The parity coverage is quite thin (3 cases), and none of them exercise nested arrays, single-row tabular arrays, or non-conforming keys — which is exactly where the encoder/decoder diverge from the spec (see the other comments). The `toon-format/spec` repo ships language-agnostic JSON fixtures (`tests/fixtures/encode/*.json`, `tests/fixtures/decode/*.json`, each with `input` / `expected` / `shouldError` / `options`). A data-driven Spock spec in this TCK that runs them through `ToonSpec` (and, in the jtoon suite, through jtoon) would give us a real conformance signal and catches regressions as the spec moves. I ran exactly that against this branch to produce the numbers in the summary — happy to share the harness.

graemerocher (changes requested):
Thanks for this — it's a substantial piece of work and the overall shape is good: a clean split between `ToonMapper`, stateless singleton encoder/decoder beans, a self-contained recursive-descent parser that reuses `LimitingStream` for the depth budget, a shared TCK with a jtoon runner mirroring how `serde-tck` is structured, and the hardening in the decoder (strict UTF-8, no pre-sizing from untrusted counts, depth charging for field lists) is exactly the kind of thing that's easy to miss. Nice.

To review this I ran the official TOON conformance fixtures from [`toon-format/spec`](https://github.com/toon-format/spec/tree/main/tests/fixtures) (spec v4.1, `tests/fixtures/{encode,decode}/*.json`) against `ToonEncoder` / `ToonDecoder` on this branch:

- **Encode: 171 pass / 7 fail**
- **Decode (strict): 311 pass / 24 fail** (a further 24 `strict: false` fixtures skipped — there is no lenient mode)

That's a strong result for a from-scratch implementation. The failures cluster into a handful of root causes, which I've left as inline comments. Two of them produce output that the reference `jtoon` decoder rejects, so I'd treat those as blocking; the rest are spec deviations / robustness that I'm fine seeing as follow-ups if you'd rather keep this PR moving.

**Blocking — encoder output not readable by jtoon 2.0.4**
1. A bare array as a list item (`- [N]:`) writes its children one indent level too deep. jtoon: `Array length mismatch: declared 2, found 0`. Any `List<List<Bean>>` / `List<List<Map>>` hits this.
2. Tabular form is used for arrays in list-item position (`- [2]{x}:`), which §9.3 / §6 forbid. jtoon: `Keyless array header with field list only valid at document root`.

**Spec deviations — encode**
3. Tabular eligibility requires ≥ 2 elements; §9.3 has no minimum for arrays (only §9.5 keyed objects require ≥ 2). Single-row arrays should encode as e.g. `items[1]{id,name}:`.
4. Tabular eligibility requires identical key *order*; §9.3 / §9.5 require the same key *set*, with header order taken from the first element. Mostly affects `Map` payloads, but it is the canonical form other encoders (jtoon included) produce.

**Spec deviations — decode**
5. §7.4 requires decoders to accept any unquoted key token up to the first unquoted `:` / `[` (e.g. `foo-bar: 1`, `2key: x`, `Content-Type: ...`) and any field name in a field list; the parser rejects them. Given that a major TOON use case is LLM-produced input, this one matters in practice — jtoon accepts them.
6. A keyed entry row with nothing after the key (`a:`) decodes to `{v: ""}` instead of erroring.
7. Strict-mode leniencies: blank lines inside array bodies and non-multiple / depth-jumping indentation are accepted (spec: MUST error; jtoon errors on both). Lower priority — being lenient on input is defensible — but the class javadoc says it operates in strict mode, so either tighten or reword.
8. Lone surrogates (`\uD800`) are accepted (§7.1 says reject), and `\u` hex parsing accepts a sign.

**API / integration**
9. `ToonMapper` doesn't override `cloneWithConfiguration` / `cloneWithViewClass` / `updateValue*`, so `ObjectMapper.cloneWithConfiguration(...)` silently returns the same instance with the configuration ignored, `@JsonView` isn't honoured, and the configured coercion policy isn't applied. `YamlObjectMapper` on this branch is a good template for all of these.
10. No HTTP `MessageBodyHandler` for `text/toon` (the spec's provisional media type), unlike `serde-yaml`. Fine as a follow-up, but let's open an issue so it isn't lost.

**Testing suggestion**
The parity spec is thin (3 cases). The spec fixture suite is plain JSON and language-agnostic; wiring it into `serde-toon-tck` as a data-driven spec would have caught every item above and keeps us honest as the spec evolves. Happy to share the throwaway Spock harness I used.

Minor nits inline (docs snippet convention, `@Internal` on the public configuration class).

realbucksavage:
@graemerocher all items from the converstaion are now addressed.

PR #1428, opened 2026-09-18
fix(support): deserialize an explicit empty object into a non-null nullable bean

### LINKED ISSUE #1411: >=3.1.0 Breaks deserializing an empty object into one that has all empty keys
### Expected Behavior

Sending an empty JSON object `{}` in a class that has all nullable (with null defaults) should yield a non-null instance of said class with all the properties set to null.

### Actual Behaviour

Starting in 3.1.0 and higher this fails with an error:
```
io.micronaut.serde.exceptions.SerdeException: Null value encountered during deserialization of type: UserInfo userInfo
	at io.micronaut.serde.support.deserializers.SpecificObjectDeserializer.requireNonNull(SpecificObjectDeserializer.java:98)
	at io.micronaut.serde.support.deserializers.SpecificObjectDeserializer.deserialize(SpecificObjectDeserializer.java:76)
	at io.micronaut.serde.Deserializer.deserializeNullable(Deserializer.java:86)
```

This is seen when attempting to deserialize a [UserInfo](https://github.com/tmapes/mn-serde-empty-json-null-repro/blob/main/src/main/kotlin/com/example/DataObjects.kt#L11) object that's nested inside another [object](https://github.com/tmapes/mn-serde-empty-json-null-repro/blob/main/src/main/kotlin/com/example/DataObjects.kt#L9)

### Steps To Reproduce

1. Clone example repo
2. Ensure `serdeVersion` is set to `brokenAfterSerde` in `build.gradle.kts`
3. Run all tests in `DataObjectsTest.kt`
4. `testSerde_UserInfo_Nested` should fail
5. Set `serdeVersion` is set to `workingSerde` in `build.gradle.kts` & sync Gradle
6. Run all tests in `DataObjectsTest.kt`
7. All tests should pass

### Environment Information

OS: macOS + Linux\
JDK: 25\
Micronaut Serde Jackson >= 3.1.0
Micronaut v5

### Example Application

https://github.com/tmapes/mn-serde-empty-json-null-repro

### Version

5.1.13

We first noticed this error when we upgraded our platform version from `5.0.4` to `5.1.2` and it was eventually traced to some change in serialization 3.1.0

mkulawinekvonage:
Fixes #1411

## Problem

Since 3.1.0, deserializing an explicit empty object (`{}`) into a *nested nullable* bean throws:

```
io.micronaut.serde.exceptions.SerdeException: Null value encountered during deserialization of type: UserInfo userInfo
```

A top-level `{}` deserializes fine; only a nullable property fails.

`ArgsConstructorBeanDeserializer.provideInstance` collapses an all-null bean to `null` when the argument is nullable:

```java
if (objectArgument.isNullable() && allNull(values) && propertiesConsumer == null && anyValuesDeserializer == null) {
    return null;
}
```

That branch was added for `@JsonUnwrapped` properties, which are reconstructed from the enclosing object's keys and have no object of their own. But `provideInstance` is also the normal decode path, so an explicit `{}` for a nullable property returned `null`, which then trips the `requireNonNull` guard in `deserialize(...)`. An explicit `{}` is not `null`: it must yield a non-null instance with all-null properties, matching 3.0.0, Jackson Databind, and the `SimpleRecordLikeObjectDeserializer` fast path.

## Fix

`provideInstance` takes a new `allowNullInstance` flag. The all-null &rarr; `null` collapse now applies only when it is set, which is only the `@JsonUnwrapped` call sites. The regular decode path uses the flag-less overload (`false`), so an actually-decoded object is never collapsed to `null`.

## Tests

`DeserializeSpec` "deserialize a nullable constructor-only property with all null values" had codified the buggy behavior (expecting the exception); it now asserts a non-null instance with a null property. The `@JsonUnwrapped` / `@JsonInclude` behavior the branch was added for is unchanged.

Verification (JDK 25): `:micronaut-serde-support:test`, `:micronaut-serde-tck:test`, `:micronaut-serde-jackson:test` (495 tests) &mdash; 0 failures; `checkstyleMain`, `spotlessCheck` and `japiCmp` pass.


mkulawinekvonage:
Closing: unable to sign the Contributor License Agreement.

PR #1429, opened 2026-09-18
fix(support): deserialize an explicit empty object into a non-null nullable bean

### LINKED ISSUE #1411: >=3.1.0 Breaks deserializing an empty object into one that has all empty keys

iwan41:
Fixes #1411

## Problem

Since 3.1.0, deserializing an explicit empty object (`{}`) into a nested nullable bean throws:

```
io.micronaut.serde.exceptions.SerdeException: Null value encountered during deserialization of type: UserInfo userInfo
```

A top-level `{}` is fine; only a nullable property fails. `ArgsConstructorBeanDeserializer.provideInstance` collapses an all-null nullable bean to `null`. That is correct for `@JsonUnwrapped` (no dedicated object in the input), but the same path also handles a normal decode, so an explicit `{}` became `null` and hit `requireNonNull`.

## Fix

Gate the all-null → `null` collapse behind `allowNullInstance`, set only at `@JsonUnwrapped` call sites. The regular decode path keeps a non-null instance with all-null properties.

## Tests

`DeserializeSpec` now asserts a non-null instance with a null property for an explicit `{}`. Existing `@JsonUnwrapped` / `@JsonInclude` coverage is unchanged.

Verification (JDK 25): `:micronaut-serde-support:test`, `:micronaut-serde-tck:test`, `:micronaut-serde-jackson:test`, `cM`, `spotlessCheck`, `japiCmp`.

Made with [Cursor](https://cursor.com)


dstepanov:
Thanks @iwan41 for the fix! Superseded by #1437, which keeps your commit (with authorship) and adds end-to-end regression tests plus a related fix for the setter-path `@JsonUnwrapped` nullability check.

